PR #22, opened 2022-06-02
dark mode console, symbols, new logo, corrections

mjkzy:
logo.bmp resource isn't working, pretty sure it's the way i'm loading it or it's not a resource?


momo5502 (approved):

src/client/component/console.cpp
```diff
@@ -45,29 +36,135 @@ namespace console
...
+			SetWindowTextA(*game::s_wcd::hwndBuffer, cleanConsoleBuffer);
+		}
 	}
 
 	class component final : public component_interface
 	{
 	public:
 		void post_unpack() override
 		{
+			const utils::nt::library self;
```
momo5502, line 154: this is not self. use const auto self = utils::nt::library::get_by_address(sys_create_console_stub);

PR #44, opened 2022-10-05
Rawfile dumping overriding

JezuzLizard:


src/client/component/assets.cpp
```diff
@@ -0,0 +1,87 @@
...
+				char* buffer = rawFile->buffer;
+				int numBytesRead = game::FS_ReadFile(rawFile->name, reinterpret_cast<void**>(&buffer));
+				if (numBytesRead > -1)
+				{
+					printf("Overriding rawfile %s\n", rawFile->name);
+					memcpy(reinterpret_cast<void*>(rawFile->buffer), buffer, numBytesRead);
+					rawFile->len = numBytesRead;
+					game::FS_FreeFile(buffer);
+				}
+			}
```
momo5502, line 46: How is that safe?
You're storing the data in the struct and replacing the buffer pointer. Why memcpy again then? You're memcpy'ing from and to the identical buffer, or am I wrong?
If not, it's even worse, cause then you don't know the size of the buffer you are copying to. So that's either a buffer overflow or a use-after-free after FS_FreeFile. No matter which case of the two, this definitely needs to be reworked to be memory safe.

momo5502 (changes requested):
Memory safety needs to be reworked. Rest is fine tho.

src/client/component/assets.cpp
ineedbots: it isn't the same buffer that memcpy is copying to, however, this isnt memory safe,  you need to add some sort of buffer pool and copy the rawfile data into it, and use that for rawfile->buffer

momo5502: Yeah, true, just noticed the result buffer is stored on the `buffer` stack variable. Memcpy'ing in this case is a buffer overflow.

JezuzLizard:
https://github.com/momo5502/boiii/pull/44/commits/bacdce179845fb706e7fd60ea5f0fc2e8eeb2084 changes rawfile overriding to store the rawfile buffers, and resize them before memcpying. Also an include for unused header file was removed.

src/client/game/structs.hpp
```diff
@@ -444,4 +444,147 @@ namespace game
 		ERR_LOCALIZATION = 6,
 		ERR_MAPLOADERRORSUMMARY = 7,
 	};
+
+	enum XAssetType : __int32
```
diamante0018, line 448: I think you should remove the `: __int32` it's from IDA, it's not necessary (int32_t is the default value anyway) and a Microsoft-specific extension.

diamante0018 (changes requested):

src/client/component/assets.cpp
```diff
@@ -0,0 +1,99 @@
...
+			//Name and buffer can be blank
+			if (rawFile->name && rawFile->buffer)
+			{
+				printf("Dumping rawfile %s\n", rawFile->name);
+				utils::io::write_file("dump\\"s + rawFile->name, rawFile->buffer);
+			}
+		}
+	}
+
+	void override_rawfile(game::RawFile* rawFile)
```
diamante0018, line 30: Just return early. forcing yourself to write functions with just one return path (end of functions) can be bad and make code harder to read.
```
if (!raw_file)
   return;
   ```

Comment (diamante0018) [inline src/client/component/assets.cpp:30]: snake case all variables. I think this is how @momo5502 has styled this project

**Comment by JezuzLizard** [comment]
https://github.com/momo5502/boiii/pull/44/commits/ccf76aaa29860ac12143936de5baa6664d844548
Changes made:

Add comments, convert all variables to snake_case.

Dumping:
Dump null buffered rawfiles, don't dump referenced rawfiles.

Fixed "Expression: vector subscript out of range" assertion.

Allow overriding null buffer rawfiles.

File: src/client/component/assets.cpp [hunk 4]
```diff
@@ -0,0 +1,134 @@
...
+
+		// Copy buffer
+		std::string buffer{};
+		buffer.resize(rawfile->len);
+		if (rawfile->buffer) // Some rawfiles such as gamedata\configs\zm\zm_reset_stats_online.cfg have null buffer
+		{
+			memcpy(&buffer[0], rawfile->buffer, rawfile->len);
+		}
+
+		utils::io::write_file("dump\\"s + rawfile->name, buffer);
```
momo5502, line 42: std::string buffer(rawfile->buffer, rawfile->len);

You can directly construct an std::string, no need to use memcpy. Just the nullcheck needs to be done

src/client/component/assets.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+			// Make sure buffer is null terminated
+			rawfile_buffer_iterator->second[num_bytes_read] = '\0';
+
+			// Assign values to asset
+			rawfile->len = num_bytes_read;
+			rawfile->buffer = reinterpret_cast<char*>(&rawfile_buffer_iterator->second[0]);
+
+			// Free the buffer
+			game::FS_FreeFile(buffer);
```
momo5502, line 93: This indeed fixes the buffer overrun. I think std::map iterators are stable, which means taking their pointer is stable as an insertion does not invalidate said pointer.
However, erasure above seems quite unsafe. Did you assert that the old pointer is definitely unused anymore? E.g. if you are overriding an asset, the old asset may still point to the old element, causing a use after free. Or can potential race conditions occur, eg. through other threads caching/simultaneously accessing the buffer?

src/client/component/assets.cpp
```diff
@@ -0,0 +1,134 @@
...
+			}
+			override_rawfile(header.rawfile);
+			break;
+		case game::ASSET_TYPE_STRINGTABLE:
+			break;
+		case game::ASSET_TYPE_SCRIPTPARSETREE:
+			break;
+		default:
+			break;
+		}
```
momo5502, line 120: Do we actually need the other cases, except rawfile?

JezuzLizard:
New version
Use unipue_ptr instead of vector for rawfile_buffers.
Dump assets by zone filename.

src/client/component/assets.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+		if (rawfile->name[0] == ',')
+		{
+			return;
+		}
+
+		char* buffer{};
+		auto num_bytes_read = game::FS_ReadFile(rawfile->name, reinterpret_cast<void**>(&buffer));
+		
+		if (num_bytes_read > -1)
```
diamante0018, line 55: `if (num_bytes_read < 0) return;` !

src/client/component/assets.cpp
```diff
@@ -0,0 +1,104 @@
...
+			}
+			rawfile_buffers[rawfile->name] = std::make_unique<std::uint8_t[]>(num_bytes_read + 1);
+			rawfile_buffer_iterator = rawfile_buffers.find(rawfile->name);
+			auto rawfile_entry_buffer = rawfile_buffer_iterator->second.get();
+
+			std::memcpy(rawfile_entry_buffer, buffer, num_bytes_read);
+			rawfile_entry_buffer[num_bytes_read] = '\0';
+
+			rawfile->len = num_bytes_read;
+			rawfile->buffer = reinterpret_cast<char*>(rawfile_entry_buffer);
```
diamante0018, line 72: change the type of the container from uint8_t to char if you need to store characters instead of doing weird conversions maybe?
You defined rawfile struct as having one buffer field of type char.

diamante0018 (changes requested):
Jesus requested me to review this pr so I'm gonna point out everything that bothers me.
A bit of changes before it's merged should be done

src/client/component/assets.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+	// TODO: Dump rawfile into parent zone folder
+	void dump_rawfile(game::RawFile* rawfile)
+	{
+		if (!rawfile || !rawfile->name || rawfile->name[0] == '\0')
+		{
+			return;
+		}
+		
+		if (rawfile->name[0] == ',')
```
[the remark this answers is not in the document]
JezuzLizard: Checking for the comma prevents attempting to dump a referenced asset.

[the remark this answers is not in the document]
diamante0018: Good to know. I guess if this is something that is commonly known it does not need a comment but I had no idea so it might not be obvious

src/client/game/structs.hpp
```diff
@@ -444,4 +444,273 @@ namespace game
 		ERR_LOCALIZATION = 6,
 		ERR_MAPLOADERRORSUMMARY = 7,
 	};
+
+	enum XAssetType
```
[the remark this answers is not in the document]
JezuzLizard: This enum is correct and can be verified by going to this address in boiii (0x14332A080). This is g_assetNames and the asset names are ordered in the same way as the enum up to 107. The cod_debug.elf's XAssetTypes enum is only missing two assets otherwise its correct. I already added the missing two asset values so the whole enum matches what can be seen here:
image

src/client/component/assets.cpp
```diff
@@ -0,0 +1,104 @@
...
+		{
+			return;
+		}
+
+		printf("Dumping rawfile %s bytes: %d\n", rawfile->name, rawfile->len);
+		std::string buffer{};
```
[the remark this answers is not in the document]
momo5502: i like {} tho, doing that everywhere leads to less uninitialized variables. even though it is unnecessary here, it is good practice imo, if you do it everywhere

src/client/component/assets.cpp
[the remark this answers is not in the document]
momo5502: yeah, kinda weird, but it is like that in cod :P

src/client/component/assets.cpp
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		char* buffer{};
+		auto num_bytes_read = game::FS_ReadFile(rawfile->name, reinterpret_cast<void**>(&buffer));
+		
+		if (num_bytes_read > -1)
+		{
+			printf("Overriding rawfile %s\n", rawfile->name);
+
+			auto rawfile_buffer_iterator = rawfile_buffers.find(rawfile->name);
```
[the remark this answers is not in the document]
momo5502: while i don't like abbreviations, i still aggree here, an iterator is an abstract concept, so the name is kinda meaningless anyways. therefore the shorter the name, the less bloat there is to read

src/client/component/assets.cpp
diamante0018: Update the comment as I realized you may prefer `<=0`. Not sure please double check

src/client/component/assets.cpp
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		printf("Dumping rawfile %s bytes: %d\n", rawfile->name, rawfile->len);
+		std::string buffer{};
+		if (rawfile->buffer)
+		{
+			buffer = std::string(rawfile->buffer, rawfile->len);
+		}
+
+		utils::io::write_file("dump\\"s + game::g_load->filename + "\\"s + rawfile->name, buffer);
```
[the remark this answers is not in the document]
JezuzLizard: The assert isn't needed because the game will crash before reaching my code. The only way g_load->filename can be nullptr is if the game tried to load a fastfile with no name.

src/client/game/structs.hpp
```diff
@@ -444,4 +444,273 @@ namespace game
...
+	{
+		XAsset asset;
+		char zoneIndex;
+		volatile char inuseMask;
+		unsigned int nextHash;
+		unsigned int nextOverride;
+		unsigned int nextPoolEntry;
+	};
+	
+	struct DB_LoadData
```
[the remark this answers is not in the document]
ineedbots: what good is the struct if we lose the actual offsets of key data? why would we remove the actual structure of the struct? it seems counter productive to me, especially if needing to reimport the structs.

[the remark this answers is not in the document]
ineedbots: at that point, it would be better off not using the struct at all and just use reinterpet_cast<char*>(<BASEADDR> + <MEMBEROFFSET>)

src/client/component/assets.cpp
JezuzLizard: 0 is a valid return value so < 0 is correct. Use case would be if you want to completely nullify what a .cfg file does by removing its contents.

src/client/component/assets.cpp
```diff
@@ -0,0 +1,104 @@
...
+		{
+			printf("Overriding rawfile %s\n", rawfile->name);
+
+			auto rawfile_buffer_iterator = rawfile_buffers.find(rawfile->name);
+			if (rawfile_buffer_iterator != rawfile_buffers.end())
+			{
+				rawfile_buffers.erase(rawfile_buffer_iterator);
+			}
+			rawfile_buffers[rawfile->name] = std::make_unique<std::uint8_t[]>(num_bytes_read + 1);
+			rawfile_buffer_iterator = rawfile_buffers.find(rawfile->name);
```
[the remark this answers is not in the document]
JezuzLizard: We need to use the ptr that make_unique created. So we can assign it to the rawfile struct. If it can be done another way please give an example.

src/client/component/assets.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+		char* buffer{};
+		auto num_bytes_read = game::FS_ReadFile(rawfile->name, reinterpret_cast<void**>(&buffer));
+		
+		if (num_bytes_read > -1)
+		{
+			printf("Overriding rawfile %s\n", rawfile->name);
+
+			auto rawfile_buffer_iterator = rawfile_buffers.find(rawfile->name);
+			if (rawfile_buffer_iterator != rawfile_buffers.end())
```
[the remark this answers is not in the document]
JezuzLizard: We need the iterator to manually delete the old buffer if we find it.

src/client/game/structs.hpp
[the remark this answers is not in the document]
JezuzLizard: I replaced the majority of the unlabeled struct with padding.

JezuzLizard:
Rebased with upstream. Move assets.cpp to assets folder, move rawfile specific code to rawfile.cpp and rawfile.hpp.

src/client/game/structs.hpp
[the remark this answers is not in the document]
momo5502: as the author of the commit, please don't mark conversations as resolved. the reviewer decides if it's resovled or not. otherwise you could just directly review your own stuff.
but I'm fine with such structs.

src/client/component/assets/rawfile.cpp
```diff
@@ -0,0 +1,87 @@
...
+			}
+
+			printf("Overriding rawfile %s\n", rawfile->name);
+
+			auto itr = rawfile_buffers.find(rawfile->name);
+			if (itr != rawfile_buffers.end())
+			{
+				rawfile_buffers.erase(itr);
+			}
+			rawfile_buffers[rawfile->name] = std::make_unique<char[]>(num_bytes_read + 1);
```
momo5502, line 69: i personally would use std::vector, but char[] is not necessarily wrong. can't decide what's better

momo5502: if you want just keep it that way

src/client/component/assets/rawfile.cpp
```diff
@@ -0,0 +1,87 @@
...
+			game::FS_FreeFile(buffer);
+		}
+
+		class component final : public component_interface
+		{
+		public:
+		};
+	}
+}
+REGISTER_COMPONENT(assets::rawfile::component)
```
momo5502, line 87: why is this a component tho? just move this back to assets? because this is definitely not a component, if that stuff is empty

src/client/component/assets/rawfile.cpp
momo5502: or actually std::string? std::unique_ptr<std::string>?

src/client/component/assets/rawfile.cpp
```diff
@@ -0,0 +1,87 @@
...
+			if (itr != rawfile_buffers.end())
+			{
+				rawfile_buffers.erase(itr);
+			}
+			rawfile_buffers[rawfile->name] = std::make_unique<char[]>(num_bytes_read + 1);
+			itr = rawfile_buffers.find(rawfile->name);
+			auto rawfile_entry_buffer = itr->second.get();
+
+			std::memcpy(rawfile_entry_buffer, buffer, num_bytes_read);
+			rawfile_entry_buffer[num_bytes_read] = '\0';
```
momo5502, line 74: yep, if you use std::string then you don't need this. .data() automatically appends nullterminator

momo5502 (changes requested):

src/client/component/assets/rawfile.cpp
JezuzLizard: Yeah I guess it doesn't need to be a component but I think it should be in its own file because I intend on adding more assets in the future so putting it in its own file makes sense because it prevents clutter.

src/client/component/assets/rawfile.cpp
ineedbots: isnt that what .c_str() does?  .data() is only going to be the data buffer to the std::string, so it may not have a null char

diamante0018: That is wrong for std::string
.data and c_string have had the same meaning for many c++ standards iterations

ineedbots: Assumes that the compiler is adhering to the standard, it IS left up to compiler implementation, but in the case of MSVC, they do the same thing

diamante0018: That's unjustified misplaced trust, MSVC STL implementation is here: https://github.com/microsoft/STL

you are probably looking for your non-existent non compliant implementation of std::string in the xstring header

(sorry for being blunt)

src/client/component/assets/rawfile.cpp
momo5502: but the component folder is definitely the wrong place, if it's not a component

src/client/component/assets/rawfile.cpp
momo5502: > isnt that what .c_str() does? .data() is only going to be the data buffer to the std::string, so it may not have a null char

Since C++11 both are the same ( https://en.cppreference.com/w/cpp/string/basic_string/data ). Therefore I prefer data

src/client/component/assets/rawfile.cpp
```diff
@@ -66,22 +66,15 @@ namespace assets
 			{
 				rawfile_buffers.erase(itr);
 			}
-			rawfile_buffers[rawfile->name] = std::make_unique<char[]>(num_bytes_read + 1);
+			rawfile_buffers[rawfile->name] = std::make_unique<std::string>();
```
momo5502, line 69: can resize the buffer you read into and directly move it into the make_unique construction sequence

src/client/component/assets/rawfile.cpp
```diff
@@ -66,22 +66,15 @@ namespace assets
...
 			itr = rawfile_buffers.find(rawfile->name);
-			auto rawfile_entry_buffer = itr->second.get();
+			auto rawfile_buffer = itr->second.get();
 
-			std::memcpy(rawfile_entry_buffer, buffer, num_bytes_read);
-			rawfile_entry_buffer[num_bytes_read] = '\0';
+			rawfile_buffer->assign(buffer, num_bytes_read);
```
momo5502, line 73: instead of copying here

src/client/component/assets/rawfile.cpp
JezuzLizard: Ah I see yeah I'll move it out of components. client\game\assets or client\assets then maybe.

momo5502:
The topic of this PR is sadly slightly more complicated than it seemed at first.
This PR does not meet the standarts of the project, yet.
As no actions have been taken recently, I'm closing it, but it can be reopend if changes are willing to be made.

PR #68, opened 2022-10-24
Skip un-handled reportReward task init

hosseinpourziyaie:
shitty fix for un-handled reportReward task with action name get_user_periodic_rarities; purpose unknown 

it retries for ever untill it gets valid response from server and shapes a massive dw task spam if not disabled


hosseinpourziyaie:
Call of Duty  Black Ops 3 Screenshot 2022 10 24 - 00 53 51 61

picture of task log lineup shaped within first minute from game's start (*before this fix)

PR #81, opened 2022-11-21
refactor(cmd): add support for cmds + cleanup

diamante0018:


src/client/game/game.hpp
```diff
@@ -50,6 +50,13 @@ namespace game
 	private:
 		size_t address_;
 	};
+
+	struct cmd_args_t
+	{
+		CmdArgs* operator->() const;
+	};
```
momo5502, line 57: seems like this belongs into symbols

momo5502 (changes requested):
one small thing, otherwise great contribution :)

src/client/game/symbols.hpp
```diff
@@ -84,6 +84,14 @@ namespace game
 	// Global game definitions
 	constexpr auto CMD_MAX_NESTING = 8;
 
-	// Reimplementations
+	struct cmd_args_t
+	{
+		CmdArgs* operator->() const
```
momo5502, line 89: shouldn't this be inline?

diamante0018: image

diamante0018: I can make it static inline, which would be more appropriate imo now that you make me think about it

diamante0018: Doesn't seem to make a practical difference. Up to you, if you still want it let me know.

momo5502: ay, alright :D

momo5502 (approved):

PR #83, opened 2022-11-25
refactor(memory): remove nullptr check before std::free

diamante0018:
https://en.cppreference.com/w/cpp/memory/c/free

### DIFF
--- a/src/common/utils/memory.cpp
+++ b/src/common/utils/memory.cpp
@@ -72,7 +72,7 @@ namespace utils
 
 	void* memory::allocate(const size_t length)
 	{
-		return calloc(length, 1);
+		return std::calloc(length, 1);
 	}
 
 	char* memory::duplicate_string(const std::string& string)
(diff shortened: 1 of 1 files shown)


src/common/utils/memory.cpp
```diff
@@ -84,22 +84,19 @@ namespace utils
...
+		std::free(data);
 	}
 
 	void memory::free(const void* data)
 	{
 		free(const_cast<void*>(data));
 	}
 
-	bool memory::is_set(const void* mem, const char chr, const size_t length)
+	bool memory::is_set(const void* mem, const char chr, const std::size_t length)
```
momo5502, line 95: what's the point of all those stds?

diamante0018: I copied and pasted in a file from a project of mine so I didn't have to manually do this on every memory.cpp 😳

momo5502 (approved):

PR #86, opened 2022-11-28
Unlocked blackmarket loot, Blackjack specialist, mega gobblegums, extra CaC slots, CWL camo's and paid specialist outfits

JariKCoding:


src/client/component/loot.cpp
```diff
@@ -0,0 +1,62 @@
+#include <std_include.hpp>
+#include "loader/component_loader.hpp"
+
+#include <utils/hook.hpp>
+#include <utils/thread.hpp>
+
+#include <component/scheduler.hpp>
```
diamante0018, line 7: Only `schduler.hpp` will work

src/client/component/loot.cpp
```diff
@@ -0,0 +1,62 @@
+#include <std_include.hpp>
+#include "loader/component_loader.hpp"
+
+#include <utils/hook.hpp>
+#include <utils/thread.hpp>
+
+#include <component/scheduler.hpp>
+#include "game/game.hpp"
```
diamante0018, line 8: Usually, in other momo's project `game` header is put at the top not bottom (after `#include "loader/component_loader.hpp"`)
The reason is that when you include other components that require game definition you will get compilation error if game.hpp is not included at the top.

diamante0018 (commented):
Very interesting pr 😀

src/client/component/loot.cpp
momo5502: While this is true, the reason is not entirely correct. The reason this is the case is that I don't really have a header preference. Ideally, each header should compile on its own, meaning it whould be irrelevant if you include game before or after other headers.

momo5502 (approved):

src/client/component/loot.cpp
JariKCoding: Thanks for the feedback! I'll keep that in mind in the future.

PR #106, opened 2023-01-09
fix infinite loop

mjkzy:


src/client/component/patches.cpp
```diff
@@ -0,0 +1,32 @@
...
+{
+	namespace
+	{
+		void scr_are_textures_loaded_stub([[maybe_unused]] game::scriptInstance_t inst)
+		{
+			game::Scr_AddInt(game::SCRIPTINSTANCE_SERVER, 1);
+		}
+	}
+
+	struct component final : generic_component
```
diamante0018, line 19: `struct component final : server_component`

diamante0018 (changes requested):

src/client/component/patches.cpp
```diff
@@ -0,0 +1,32 @@
+#include <std_include.hpp>
+
```
diamante0018, line 2: remove newline to fix ocd

src/client/component/patches.cpp
diamante0018: if you don't this will crash when loading the client because it's a generic component

diamante0018: never mind, say is_server check

mjkzy: so is it fine? tbh its more of a future thing but if its a big deal then i can change :p

diamante0018: Consider renaming patches.cpp to dedi_patches.cpp and for client patches maybe momo can later down the road client_patches. otherwise this is fine

mjkzy: i hate this new style >:( 
will change

diamante0018 (approved):

momo5502 (approved):

Issue #158, opened 2023-02-13
Double click a server name to join


ethanjones99:
Just a feature request and it would be a nice QoL improvement.

rec0veryyy:
I support this idea, double click instead of just pressing the J for uniser.

PR #186, opened 2023-02-16
feat: add 'double-click to join server' to server browser

Lierrmm:
Double-Click Server in Server Browser to Join.

resolves momo5502/boiii#158

### DIFF
--- a/data/ui_scripts/doubleclick_join_server/__init__.lua
+++ b/data/ui_scripts/doubleclick_join_server/__init__.lua
@@ -1,0 +1,113 @@
+require( "ui.uieditor.widgets.PC.ServerBrowser.ServerBrowserRowInternal" )
+require( "ui.uieditor.widgets.Lobby.Common.FE_FocusBarContainer" )
+
+CoD.ServerBrowserRow = InheritFrom( LUI.UIElement )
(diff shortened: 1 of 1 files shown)


data/ui_scripts/doubleclick_join_server/__init__.lua
```diff
@@ -0,0 +1,113 @@
...
+	self.soundSet = "default"
+	self:setLeftRight( true, false, 0, 700 )
+	self:setTopBottom( true, false, 0, 22 )
+	self:makeFocusable()
+	self:setHandleMouse( true )
+
+	if CoD.isPC then
+		self:registerEventHandler( "leftmousedown", function ()
+			if self:isInFocus() then
+				JoinServerBrowser( self, self, controller, menu )
```
JariKCoding, line 22: This will also need a call to `ServerBrowserCancelRequest( self )` or else you will make multiple join requests when clicking more that 2 times / when reopening the server browser. This is also done in other scripts where JoinServerBrowser is called.

JariKCoding (changes requested):
Just needs 1 small change, other than that it looks good.

momo5502 (approved):
trusting @JariKCoding here.

PR #187, opened 2023-02-16
Change server console window title to server name

### LINKED ISSUE #164: Process Title
Would you be able to add the ability to change the process name of the dedicated servers so we know what servers to close if there are errors? At the moment all my processes are called "BOIII Server" and I dont know what server is what.

Brentdevent:
Resolves #164

### DIFF
--- a/src/client/component/console.cpp
+++ b/src/client/component/console.cpp
@@ -3,6 +3,7 @@
 #include "resource.hpp"
 
 #include "game/game.hpp"
+#include "scheduler.hpp"
 
 #include <utils/thread.hpp>
 #include <utils/hook.hpp>
@@ -203,6 +204,14 @@ namespace console
 			if (!game::is_server())
 			{
 				utils::hook::set<uint8_t>(0x14133D2FE_g, 0xEB); // Always enable ingame console
+			}
+			else
+			{
+				scheduler::once([]()
+				{
+					const auto server_name = game::Dvar_FindVar("live_steam_server_name")->current.string;
+					SetWindowTextA(*game::s_wcd::hWnd, server_name);
+				}, scheduler::pipeline::main);
 			}
 
 			utils::hook::jump(game::select(0x1423337F0, 0x1405976B0), queue_message);
--- a/src/client/component/console.cpp
+++ b/src/client/component/console.cpp
@@ -1,4 +1,5 @@
 #include <std_include.hpp>
+#include "console.hpp"
 #include "loader/component_loader.hpp"
 #include "resource.hpp"
 
@@ -15,6 +16,11 @@
 
 namespace console
 {
+	void set_title(const std::string& title)
+	{
+		SetWindowTextA(*game::s_wcd::hWnd, title.data());
+	}
+
 	namespace
 	{
 		utils::image::object logo;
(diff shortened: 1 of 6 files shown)


src/client/component/console.cpp
```diff
@@ -204,6 +205,14 @@ namespace console
 			{
 				utils::hook::set<uint8_t>(0x14133D2FE_g, 0xEB); // Always enable ingame console
 			}
+			else
+			{
+				scheduler::once([]()
+				{
+					const auto server_name = game::Dvar_FindVar("live_steam_server_name")->current.string;
```
diamante0018, line 212: Is this dvar always guaranteed to be registered, in other words, no nullptr check in ret val of Dvar_FindVar?

Brentdevent: Pretty sure it gets registered everytime, but will add a check just to be sure 👍🏼

src/client/component/console.cpp
```diff
@@ -204,6 +205,17 @@ namespace console
...
+			else
+			{
+				scheduler::once([]()
+				{
+					const auto server_name_dvar = game::Dvar_FindVar("live_steam_server_name");
+					if (server_name_dvar)
+					{
+						SetWindowTextA(*game::s_wcd::hWnd, server_name_dvar->current.string);
+					}
+				}, scheduler::pipeline::main);
```
momo5502, line 217: please don't do this every frame, do it only every second or so

this could also be extend to player count/max players

momo5502 (changes requested):

Brentdevent:
I tried adding the bot count to the infoResponse as well:

https://github.com/momo5502/boiii/blob/29095b0224ed2a0b172192e1a30dcf9c712cd781/src/client/component/getinfo.cpp#L89

But for some reason it disconnected me with some error code.

src/client/game/symbols.hpp
```diff
@@ -27,7 +27,7 @@ namespace game
 	};
 
 	WEAK symbol<void(uint32_t localClientNum, const char* text)> Cbuf_AddText{0x1420EC8B0, 0x1404F75B0};
-	WEAK symbol<void(int localClientNum, ControllerIndex_t controllerIndex, const char* buffer)> Cbuf_ExecuteBuffer{
+	WEAK symbol<void(uint32_t localClientNum, ControllerIndex_t controllerIndex, const char* buffer)> Cbuf_ExecuteBuffer{
```
diamante0018, line 30: I do tho think these should be a signed integer and not unsigned although it wouldn't make sense for the client number to be negative. So I'm split on this one. Momo probably knows better

Brentdevent: Yep, I was not sure myself. I'll change if necessary.

momo5502: the original type is an enum, of which the underlying type is an int, so i'd prefer int

src/client/component/dedicated_info.cpp
```diff
@@ -0,0 +1,42 @@
...
+	class component final : public server_component
+	{
+	public:
+		void post_unpack() override
+		{
+			scheduler::loop([]()
+			{
+				const auto sv_running = game::Dvar_FindVar("sv_running");
+
+				if (sv_running && sv_running->current.enabled)
```
momo5502, line 23: change this to an early return please, the nesting is quite deep here already

momo5502: also if the server is not running, reflecting that in the title might make sense, instead of displaying the previous info, doesn't it?

momo5502: maybe extracting everything in a function as well might be cleaner

momo5502 (changes requested):

src/client/component/dedicated_info.cpp
Brentdevent: Yep, you're correct will change it 👍🏼

momo5502 (approved):
Formatting seems off, but i'll fix it myself

PR #190, opened 2023-02-17
Fix line endings in scheduler

diamante0018:
they were mixed! i put them back to the default crlf


src/client/component/bots.cpp
```diff
@@ -25,6 +25,12 @@ namespace bots
 				{"momo5502", "IW5x"},
 				{"Maurice", "IW5x" },
```
momo5502, line 26: Please remove maurice 😂 It was there to test shuffeling but as momo is in there, you can yeet it

momo5502 (changes requested):

momo5502 (approved):

Issue #151, opened 2023-02-12
Bot count needs to be implemented in getinfo


JoelColby:
Pretty sure this one is already on your radar from the comments, but mainly making this issue for tracking.

`bots` is currently hard-coded to 0 as part of the `infoResponse` in: https://github.com/momo5502/boiii/blob/f70ed6dafb73d66c41979250184704281b2db226/src/client/component/getinfo.cpp#L74

To complete: need to implement `get_bot_count()`, optionally may wish to subtract bots from clients in certain places for displaying a "real" player count.

PR #201, opened 2023-02-18
feat: added bot count to getinfo

JariKCoding:
Resolves issue #151
Seems like @Brentdevent already implemented the count but didn't enable it in getinfo.


Brentdevent:
It gave me some issues: https://github.com/momo5502/boiii/pull/187#issuecomment-1435167155
That's the reason I did not enable it.

JariKCoding:
@Brentdevent Were you able to reliably reproduce it? It works fine for me and I'm not getting that error.

Brentdevent:
Hmm weird enough neither do I anymore. I'm pretty sure I double checked it. Very strange 😅 
Anyway the error I was getting seems related to the stats hook anyway, So maybe it was something else. 🤔

momo5502 (approved):
I guess we'll see if it causes any bugs :D

PR #216, opened 2023-02-21
fix: iw4m compat

diamante0018:


momo5502 (approved):

momo5502 (changes requested):
wait, what about the client?

diamante0018:
why would the client need this? I mean ok

diamante0018:
@momo5502 done

momo5502 (approved):

momo5502:
> why would the client need this? I mean ok

Why would a client need a working netadr to string implementation?
Well, right now, it's not needed, but first thing that comes to my mind for example is a command that prints the server you are connected to. or maybe the scoreboard info thingy iw4x has, which shows the server you are connected to on the scoreboard. there are plenty of reasons one might need netadr to string. dumbest reason might even be debugging

PR #222, opened 2023-02-23
feat(QoL): remove colour prefixes from server name when updating console title

Lierrmm:
Quality of Life improvement that removes coloured prefixes from server name when updating console title.

Old:
image

New:
image


src/client/component/dedicated_info.cpp
```diff
@@ -18,16 +19,20 @@ namespace dedicated_info
 			const auto sv_running = game::Dvar_FindVar("sv_running");
 			const auto server_name = game::get_dvar_string("live_steam_server_name");
 
+			// QoL improvement - Remove colour codes from server name for console title
```
momo5502, line 22: please remove the comment

image

src/client/component/dedicated_info.cpp
```diff
@@ -18,16 +19,20 @@ namespace dedicated_info
...
+			std::regex vowel_re("\\^[0-9;:]");
+			const auto clean_server_name = std::regex_replace(server_name, vowel_re, "");
```
momo5502, line 24: the game provides some I_CleanStr function which does exactly that but in a way that is guaranteed to clean the string in a compatible way, please use that instead

momo5502 (changes requested):

momo5502:
no need to close it, it was a good pr, if you strugglenwith the i cleanstr thing, i can help you

Lierrmm:
Had some trouble with some git merges so will re-open once I've resolved those.

Issue #136, opened 2023-02-10
Can't use custom classes on servers.


NotAdamGamingLOL:
Self explanatory. When I try to use one of the classes I made(for public matches) it just load ones of the starter guns like the KN44, BRM, and etc. but not my class setup. Hope this get fix before release because I really would love to use my Galil and Gorgon.

Dss0:
workaround: change the gametype to tdm and then change it back:
`gametype tdm; map_restart`
`gametype gun; map_restart`

SadSlothXL:
Also wanted to add that you cant use a different specialist you have selected, you are always force to use ruin.

rhobiusvoid:
![image](https://user-images.githubusercontent.com/22803374/220222019-ee795c55-a828-404e-900c-e2ebddb8a311.png)
The 5 default classes that the game lets you play with before you reach Rank 4 (which unlocks Create-a-Class) seem to be empty from the menu, and load up as having nothing (fists, no nades or perks) in a match. These classes can actually be changed in the Custom Games "Edit Game Rules" menu under "Create-a-Class."
![image](https://user-images.githubusercontent.com/22803374/220222199-a8840594-dd14-4696-8409-3749715199f7.png)
![image](https://user-images.githubusercontent.com/22803374/220222225-444ba461-1754-494d-92f3-bef829b10039.png)
This also suggests that the game thinks you're in a Custom Games lobby as opposed to a Public Match one. The bots can use weapons from these classes seemingly without any problems, but it's likely they have their own coding to handle their loadouts.

PR #223, opened 2023-02-23
Online dedicated servers

Brentdevent:
This fixes issue #136


momo5502 (approved):
thank god the should transfer hook is gone. great addition. thank you for that :)

PR #224, opened 2023-02-23
Set same modes as server when connecting

Brentdevent:
Fixes the wrong preview of classes which can occur when you switch to private match before joining the server.


src/client/component/party.cpp
```diff
@@ -208,6 +208,9 @@ namespace party
 				return;
 			}
 
+			game::Com_SessionMode_SetNetworkMode(game::MODE_NETWORK_ONLINE);
+			game::Com_SessionMode_SetGameMode(game::MODE_GAME_MATCHMAKING_PLAYLIST);
```
momo5502, line 212: Is it sure this will always be the right mode? What if you connect via P2P to a client that is running zombies or so in a non-playlist match?

Brentdevent: Ah yes, you're right, should have thought about that. Closing the PR for now.

Issue #225, opened 2023-02-23
Class selection changes based on previous menu before joining server


rhobiusvoid:
As the title says, the classes you're presented with when you join a team upon entering a server changes based on what menu you were in previously. It defaults to the Custom Games classes if you don't enter the Public Match menu, and it also changes if you're in the Arena menu, which also has the side-effect of preventing you from switching teams (which thus prevents you from joining any when you join a server).

However, this seems to be nothing more than a cosmetic problem, since the game will pull from the classes you set up in the Public Match screen.
![image](https://user-images.githubusercontent.com/22803374/220979380-75005008-7543-45cb-a1c3-7a5a78250ed2.png)
Up above, is the 2nd class I have in Custom Games.
![image](https://user-images.githubusercontent.com/22803374/220979452-3ced1ec8-7804-439d-a157-abc216c2451f.png)
But as shown here, it spawns in with the 2nd class I made for Public Matches, this M16 with those attachments.

rhobiusvoid:
No longer happens after pull request #226 fixed. Previous menu was Custom Games before joining server, shows Public Match classes.
![image](https://user-images.githubusercontent.com/22803374/221028956-e8998197-8c9b-42b2-87de-f961819bf91b.png)
Closing issue as fixed. Great job!

PR #226, opened 2023-02-23
Fixes zombies intro & client sided issues with dedicated servers

Brentdevent:
- This should fix issues with dedicated servers rotating through different maps because players are forced to watch the intro during map rotation.
- Classes should now be previewed correctly all the time.
- Choosing different gametypes on dedicated servers / private matches, should no longer cause any issues anymore.


momo5502 (approved):

Brentdevent:
This should fix #225

PR #231, opened 2023-02-24
feat: remove color prefixes from server name in console title

Lierrmm:
removes colored prefixes from server name when updating console title.

Old:
image

New:
image

### DIFF
--- a/src/client/component/dedicated_info.cpp
+++ b/src/client/component/dedicated_info.cpp
@@ -16,8 +16,8 @@ namespace dedicated_info
 		void set_server_info_in_console_title()
 		{
 			const auto sv_running = game::Dvar_FindVar("sv_running");
-			const auto server_name = game::get_dvar_string("live_steam_server_name");
-			const auto clean_server_name = game::I_CleanStr(server_name.data());
+			auto server_name = game::get_dvar_string("live_steam_server_name");
+			auto clean_server_name = game::I_CleanStr(server_name.data());
 
 			if (!sv_running || !sv_running->current.enabled)
 			{
--- a/src/client/component/dedicated_info.cpp
+++ b/src/client/component/dedicated_info.cpp
@@ -17,17 +17,18 @@ namespace dedicated_info
 		{
 			const auto sv_running = game::Dvar_FindVar("sv_running");
 			const auto server_name = game::get_dvar_string("live_steam_server_name");
+			const auto clean_server_name = game::I_CleanStr(server_name.data());
 
 			if (!sv_running || !sv_running->current.enabled)
 			{
-				console::set_title(server_name + " - not running");
+				console::set_title(utils::string::va("%s - not running", clean_server_name));
 				return;
 			}
 
 			const auto mapname = game::get_dvar_string("mapname");
 
 			const std::string window_text = utils::string::va("%s on %s [%d/%d] (%d)",
-			                                                  server_name.data(),
+															  clean_server_name,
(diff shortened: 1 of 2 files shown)


src/client/game/symbols.hpp
```diff
@@ -131,6 +131,9 @@ namespace game
 	WEAK symbol<void(client_s* cl_0, svscmd_type type, const char* fmt, ...)> SV_SendServerCommand{0x0, 0x140537F10};
 	WEAK symbol<bool(int clientNum)> SV_IsTestClient{0x14224B5C0, 0x14052FF40};
 
+	// Utils
+	WEAK symbol<const char* (const char* str)> I_CleanStr{ 0x1422E9C10, 0x140580E80 };
```
momo5502, line 135: remove the spaces inside the brackets and it's perfect :P
```suggestion
	WEAK symbol<const char* (const char* str)> I_CleanStr{0x1422E9C10, 0x140580E80};
```

momo5502 (changes requested):

src/client/game/symbols.hpp
momo5502: question is, does this take a const char* or a char* ? i think it modifies the buffer that comes in, so maybe changing it to that would be good: 
```suggestion
	WEAK symbol<const char* (char* str)> I_CleanStr{0x1422E9C10, 0x140580E80};
```

src/client/game/symbols.hpp
```diff
@@ -131,6 +131,9 @@ namespace game
 	WEAK symbol<void(client_s* cl_0, svscmd_type type, const char* fmt, ...)> SV_SendServerCommand{0x0, 0x140537F10};
 	WEAK symbol<bool(int clientNum)> SV_IsTestClient{0x14224B5C0, 0x14052FF40};
 
+	// Utils
+	WEAK symbol<const char* (const char* str)> I_CleanStr{0x1422E9C10, 0x140580E80};
```
momo5502, line 135: As mentioned in the previous comment, i think it modifies the input:

```suggestion
	WEAK symbol<const char* (char* str)> I_CleanStr{0x1422E9C10, 0x140580E80};
```

momo5502 (changes requested):

momo5502 (approved):

src/client/component/dedicated_info.cpp
```diff
@@ -17,17 +17,18 @@ namespace dedicated_info
 		{
 			const auto sv_running = game::Dvar_FindVar("sv_running");
 			const auto server_name = game::get_dvar_string("live_steam_server_name");
```
momo5502, line 19: just remove the const here :D
```suggestion
			auto server_name = game::get_dvar_string("live_steam_server_name");
```

momo5502: these suggestions are super nice 😂

src/client/component/dedicated_info.cpp
```diff
@@ -16,18 +16,19 @@ namespace dedicated_info
 		void set_server_info_in_console_title()
 		{
 			const auto sv_running = game::Dvar_FindVar("sv_running");
-			const auto server_name = game::get_dvar_string("live_steam_server_name");
+			auto server_name = game::get_dvar_string("live_steam_server_name");
+			auto clean_server_name = game::I_CleanStr(server_name.data());
```
momo5502, line 20: that one could be const, but fuck it

momo5502 (approved):

PR #232, opened 2023-02-24
Add templates for bug reports and feature requests

JoelColby:
This configuration still allows for the creation of blank issues, but is aimed at providing guidance for users who don't regularly use GitHub to create a good bug report.

Bug reports automatically are automatically assigned the "bug" label. Feature requests are automatically assigned the "feature" label.

Some assumptions made about the operating systems supported by this project. Happy to modify these if they are incorrect.

See a live version of the issue templates at: https://github.com/JoelColby/boiii/issues/new/choose


momo5502 (approved):
This is perfect. Thank you so much :D

PR #236, opened 2023-02-25
don't make script errors fatal error

mjkzy:


src/client/component/client_patches.cpp
```diff
@@ -69,6 +69,20 @@ namespace client_patches
...
+		void script_errors_stub(const char* file, int line, unsigned int code, const char* fmt, ...)
+		{
+			char buffer[0x1000];
+
+			{
+				va_list ap;
+				va_start(ap, fmt);
+				vsnprintf_s(buffer, sizeof(buffer), _TRUNCATE, fmt, ap);
+				va_end(ap);
+			}
```
momo5502, line 82: good practice before com error, not sure if intentional, but very good to scope

src/client/component/client_patches.cpp
```diff
@@ -82,6 +96,9 @@ namespace client_patches
 			utils::hook::set(0x15AAEB254_g, mixer_open_stub);
 
 			preload_map_hook.create(0x14135A1E0_g, preload_map_stub);
+
+			// don't make script errors fatal error
+			utils::hook::call(0x1412CAC4D_g, script_errors_stub);
```
momo5502, line 101: what about the dedi?

mjkzy:
oh yeah, forgot to do dedi. will do that inna bit

src/client/component/client_patches.cpp
mjkzy: will add dedi support in a bit

src/client/component/patches.cpp
```diff
@@ -0,0 +1,36 @@
...
+			char buffer[0x1000];
+
+			{
+				va_list ap;
+				va_start(ap, fmt);
+				vsnprintf_s(buffer, sizeof(buffer), _TRUNCATE, fmt, ap);
+				va_end(ap);
+			}
+
+			game::Com_Error(game::ERROR_SCRIPT_DROP, "%s", buffer);
```
momo5502, line 22: the address for com_error for dedis is not yet added in symbols.hpp!

src/client/component/patches.cpp
```diff
@@ -0,0 +1,36 @@
...
+		}
+	}
+
+	struct component final : generic_component
+	{
+		void post_unpack() override
+		{
+			// don't make script errors fatal error
+			utils::hook::call(game::select(0x1412CAC4D_g, 0x140158EB2_g), script_errors_stub);
```
momo5502, line 31: select doesn't require the _g literal

momo5502 (changes requested):
i wish the formatting was in a separate PR, that would make it easier to review and properly separate the concerns

src/client/component/patches.cpp
mjkzy: ah poop i should have looked into this more, sorryyyy

momo5502 (approved):

PR #238, opened 2023-02-25
Lock game modes that are not installed

counter185:
This prevents an infinite loading screen caused by starting a game mode that is not installed.

image

*this does not prevent the game from crashing when joining a server for a non-installed game mode through Server Browser

### DIFF
--- a/src/client/steam/interfaces/apps.cpp
+++ b/src/client/steam/interfaces/apps.cpp
@@ -33,14 +33,18 @@ namespace steam
 		return "english";
 	}
 
+	// This is used for checking DLC too.
 	bool apps::BIsSubscribedApp(unsigned int appID)
 	{
-		return true;
+		return appID == 366842 ? std::filesystem::exists("zone/zm_common.xpak")	//zombies
+			: appID == 366841 ? std::filesystem::exists("zone/mp_common.xpak")	//multiplayer
+			: appID == 366840 ? std::filesystem::exists("zone/cp_common.xpak")	//campaign
+			: true;
 	}
 
 	bool apps::BIsDlcInstalled(unsigned int appID)
 	{
-		return true;
+		return BIsSubscribedApp(appID);	//to not repeat the same code here
 	}
 
 	unsigned int apps::GetEarliestPurchaseUnixTime(unsigned int nAppID)
--- a/src/client/steam/interfaces/apps.cpp
+++ b/src/client/steam/interfaces/apps.cpp
@@ -38,9 +38,9 @@ namespace steam
 
 	bool apps::BIsSubscribedApp(unsigned int appID)
 	{
-		static const auto has_campaign = std::filesystem::exists(::utils::nt::library{}.get_path() / "zone/cp_common.xpak");
-		static const auto has_multiplayer = std::filesystem::exists(::utils::nt::library{}.get_path() / "zone/mp_common.xpak");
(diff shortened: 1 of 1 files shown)


mjkzy:
> this does not prevent the game from crashing when joining a server for a non-installed game mode through Server Browser
doing this should be fairly simple. maybe i'll look into it anyways

src/client/steam/interfaces/apps.cpp
```diff
@@ -33,14 +33,18 @@ namespace steam
...
 	}
 
+	// This is used for checking DLC too.
 	bool apps::BIsSubscribedApp(unsigned int appID)
 	{
-		return true;
+		return appID == 366842 ? std::filesystem::exists("zone/zm_common.xpak")	//zombies
+			: appID == 366841 ? std::filesystem::exists("zone/mp_common.xpak")	//multiplayer
+			: appID == 366840 ? std::filesystem::exists("zone/cp_common.xpak")	//campaign
+			: true;
```
momo5502, line 42: two things here: i recommend checking file existance not relative to the working directory, but the host module and caching the result is a must. like so:

```suggestion
		static const auto has_zombies = std::filesystem::exists(::utils::nt::library{}.get_path() / "zone/zm_common.xpak");
		static const auto has_multiplayer = std::filesystem::exists(::utils::nt::library{}.get_path() / "zone/mp_common.xpak");
		static const auto has_campaign = std::filesystem::exists(::utils::nt::library{}.get_path() / "zone/cp_common.xpak");
		return appID == 366842 ? has_zombies
			: appID == 366841 ? has_multiplayer
			: appID == 366840 ? has_campaign
			: true;
```

that way you can even get rid of the comments, as the variable names speak for themselves. 
You will have to include the nt util tho:

`#include <utils/nt.hpp>`

src/client/steam/interfaces/apps.cpp
```diff
@@ -33,14 +33,18 @@ namespace steam
...
 	}
 
 	bool apps::BIsDlcInstalled(unsigned int appID)
 	{
-		return true;
+		return BIsSubscribedApp(appID);	//to not repeat the same code here
```
momo5502, line 47: Please remove the comment. It's obvious.

momo5502 (changes requested):
great addition. just a few minor things

src/client/steam/interfaces/apps.cpp
counter185: I changed it to use `::utils::nt::library{}.get_path()` like you said here and now it detects all of them as not installed. `get_folder()` seems to work though

counter185:
done

src/client/steam/interfaces/apps.cpp
```diff
@@ -35,12 +38,18 @@ namespace steam
 
 	bool apps::BIsSubscribedApp(unsigned int appID)
 	{
-		return true;
+		static const auto has_campaign = std::filesystem::exists(::utils::nt::library{}.get_folder() / "zone/cp_common.xpak");
+		static const auto has_multiplayer = std::filesystem::exists(::utils::nt::library{}.get_folder() / "zone/mp_common.xpak");
+		static const auto has_zombies = std::filesystem::exists(::utils::nt::library{}.get_folder() / "zone/zm_common.xpak");
```
momo5502, line 43: ofc get_folder not get_path, i'm dumb :)

momo5502 (approved):

PR #239, opened 2023-02-25
Ranked dedicated servers

Brentdevent:
This makes progression possible on dedicated servers (zombies and multiplayer).
Note that this PR will break team changing in multiplayer. Needs to be fixed with LUI.


momo5502 (approved):
hard to review the addresses, but will need testing

PR #244, opened 2023-02-27
better discord rpc

mjkzy:
- needs a way to get the amount of clients in your game (i'm not familiar with the structs)
- discord application ID needs changed


src/client/component/discord.cpp
```diff
@@ -1,34 +1,71 @@
...
+				const auto map = game::Com_GameInfo_GetMapRef(game::Dvar_FindVar("mapname")->current.string);
 
-			discord_presence.partySize = 0;
-			discord_presence.partyMax = 0;
-			discord_presence.startTimestamp = 0;
+				discord_presence.details = utils::string::va("%s on %s", game::UI_SafeTranslateString(gametype), game::UI_SafeTranslateString(map));
 
-			Discord_UpdatePresence(&discord_presence);
-		}
+				static char clean_hostname[0x80] = {0};
```
momo5502, line 45: why static char of 8? why not std::string? why static?

src/client/component/discord.cpp
```diff
@@ -1,34 +1,71 @@
...
+#include <utils/string.hpp>
+
+#include <discord_rpc.h>
+
 namespace discord
 {
 	namespace
 	{
-		void ready(const DiscordUser* /*request*/)
+		DiscordRichPresence discord_presence{};
```
momo5502, line 18: why is this a global?

src/client/component/discord.cpp
```diff
@@ -46,11 +83,16 @@ namespace discord
 			handlers.spectateGame = nullptr;
 			handlers.joinRequest = nullptr;
 
-			Discord_Initialize("1047539933922988112", &handlers, 1, nullptr);
+			Discord_Initialize("1079371545266364516", &handlers, 1, nullptr);
```
momo5502, line 86: why did you change the id? instead of requesting me to make the changes on the proper app

src/client/component/discord.cpp
```diff
@@ -63,6 +105,22 @@ namespace discord
...
+			discord_presence.largeImageKey = "boiii";
+
+			printf("Discord: Ready on %s (%s)\n", request->username, request->userId);
+			Discord_UpdatePresence(&discord_presence);
+		}
+
+		static void errored(const int error_code, const char* message)
+		{
+			printf("Discord: Error (%i): %s\n", error_code, message);
+		}
```
momo5502, line 123: why static member functions?

src/client/component/party.cpp
```diff
@@ -28,6 +28,8 @@ namespace party
 			std::chrono::high_resolution_clock::time_point query_time{};
 		};
 
+		std::string hostname{};
```
momo5502, line 31: pretty sure the game already stores this. like iw4x has it in the scoreboard server info.

src/client/component/discord.cpp
```diff
@@ -46,11 +83,16 @@ namespace discord
...
-			Discord_Initialize("1047539933922988112", &handlers, 1, nullptr);
+			Discord_Initialize("1079371545266364516", &handlers, 1, nullptr);
 
-			this->initialized_ = true;
+			scheduler::once([]
+			{
+				scheduler::once(update_discord, scheduler::pipeline::async);
+				scheduler::loop(update_discord, scheduler::pipeline::async, 5s);
+				scheduler::loop(Discord_RunCallbacks, scheduler::pipeline::async, 1s);
+			}, scheduler::pipeline::main);
```
momo5502, line 93: that's a bit weird, that can be cleaned up

src/client/component/discord.cpp
```diff
@@ -1,34 +1,71 @@
...
+#include <discord_rpc.h>
+
 namespace discord
 {
 	namespace
 	{
-		void ready(const DiscordUser* /*request*/)
+		DiscordRichPresence discord_presence{};
+
+		void update_discord()
```
momo5502, line 20: this function is quite long and has a lot of code. it can be split up

momo5502 (changes requested):

src/client/component/discord.cpp
mjkzy: this is all code from h1-mod, keep that in mind :P

also, i think its because of utils::string::strip
https://github.com/mjkzy/boiii/blob/discord/src/client/component/discord.cpp#L53

src/client/component/discord.cpp
mjkzy: idk, this is from s1x https://github.com/XLabsProject/s1x-client/blob/develop/src/client/component/discord.cpp#L152

src/client/component/discord.cpp
mjkzy: same as above https://github.com/XLabsProject/s1x-client/blob/develop/src/client/component/discord.cpp#L130

src/client/component/discord.cpp
mjkzy: same has above. https://github.com/XLabsProject/s1x-client/blob/develop/src/client/component/discord.cpp#L34
however, i agree

src/client/component/discord.cpp
diamante0018: It's not Static on xlabs and it works fine. You just have to call the discord API function before the stack is cleaned by the epilogue code at the function end. That's what I did on iw4x, iw6x, s1x

diamante0018: Why should we keep in mind this is code from h1? It shouldn't really matter where it's from

diamante0018 (commented):
While I can't blame anyone for taking inspiration from S1x, if the code can be improved it should and then s1x shall receive an upgrade to the code as well. Replying with links to the s1x code as a means to justify why code shouldn't be changed is not the right way to go forward (that's how I interpreted it). I can update the xlabs clients after this pr is finalized with the new and improved code

src/client/component/discord.cpp
Rackover, line 86: Move application ID to a named constexpr at the beginning of the file

Rackover: (most of the strings i see here - "Main menu", "Private match" could be moved to constexpr variables the same for easier edit)

Rackover (commented):
nothing more to say than what's already been said, just these tiny bit of strings lying around in code that could be put in defines/constexpr in CPP

mjkzy:
will prob reopen this later, left it as draft because i wasn't ready for reviews yet and was just having the RPC at least working but thank you guys :)

src/client/component/discord.cpp
mjkzy: how should i do this? i do it so the `update_discord` and `ready` functions can both use the variable

src/client/component/party.cpp
mjkzy: i looked into iw4x discord RPC and just how hostname stuff is handled and they do the same solution by just getting the hostname stored from the infostring

src/client/component/discord.cpp
mjkzy: how do you want me to clean this up? we only run the discord RPC stuff when the main thread is ready (which makes sense because `update_discord` calls game functions that become relevant after that) and then we just call that same function ervery 5 seconds. and then the Discord_RunCallbacks is necessary for keeping Discord up to date

Issue #245, opened 2023-02-27
using spawnBot crashes client if not host


Lierrmm:
### Bug Type

Client

### BOIII Version

v0.0.1 (Latest)

### Environment (Operating System)

Microsoft Windows 11 (64-Bit)

### Reproducibility Rate

High: The bug can (almost) always be reproduced

### Steps to Reproduce

1. Connect to a dedi server
2. run the command `/spawnBot 2`
3. Client should crash

### Expected Result

Nothing should happen.

### Actual Result

Client Crashes

### Severity/Priority

Minor

PR #246, opened 2023-02-27
handle addtestclient properly when not host

Lierrmm:
Fixes: #245

### DIFF
--- a/src/client/component/dvars.cpp
+++ b/src/client/component/dvars.cpp
@@ -65,6 +65,9 @@ namespace dvars
 				return;
 			}
 
+			const auto [beg, end] = std::ranges::remove(data, '\r');
+			data.erase(beg, end);
+
 			std::istringstream stream(data);
 			std::string line;
 


src/client/component/client_patches.cpp
```diff
@@ -13,6 +13,7 @@ namespace client_patches
 	namespace
 	{
 		utils::hook::detour preload_map_hook;
+		utils::hook::detour sv_addtestclient_hook;
```
momo5502, line 16: do you need a detour hook for that? would it be possible to i intercept all the calls to that function instead?

momo5502: cause detours can potentially break if there are rip-relative instructions within the jump to the trampoline that gets inserted, those cannot properly be reconstructed by the reconstructed original stub

momo5502: therefore using them with care and only where necessary is desired. or you make sure there are no rip-relative instructions, that also works.

Lierrmm: I've just realized that the spawnBot command is defined in bots.cpp. 

Would it make more sense to add the logic there instead?

momo5502: yes, definitely, just make sure your clientnum check is ignored for dedis

momo5502: or better, extract it into a is_host function that always returns true on dedis

PR #247, opened 2023-02-27
Fix crashing when typing in console

Brentdevent:
File contains now a carriage return after downloading, it caused some issues.

### DIFF
--- a/src/client/component/dvars.cpp
+++ b/src/client/component/dvars.cpp
@@ -65,6 +65,8 @@ namespace dvars
 				return;
 			}
 
+			data.erase(std::remove(data.begin(), data.end(), '\r'), data.end());
+
 			std::istringstream stream(data);
 			std::string line;
 
--- a/src/client/component/dvars.cpp
+++ b/src/client/component/dvars.cpp
@@ -65,7 +65,7 @@ namespace dvars
 				return;
 			}
 
-			data.erase(std::remove(data.begin(), data.end(), '\r'), data.end());
+			data.erase(std::ranges::remove(data, '\r').begin(), data.end());
 
 			std::istringstream stream(data);
 			std::string line;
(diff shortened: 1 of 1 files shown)


momo5502 (approved):

diamante0018 (commented):
std::ranges::remove, std::ranges::remove_if?

momo5502 (approved):

momo5502 (changes requested):
https://stackoverflow.com/a/64330309/8725381

src/client/component/dvars.cpp
```diff
@@ -65,7 +65,7 @@ namespace dvars
 				return;
 			}
 
-			data.erase(std::remove(data.begin(), data.end(), '\r'), data.end());
+			data.erase(std::ranges::remove(data, '\r').begin(), data.end());
```
diamante0018, line 68: If my understanding is correct this isn't right anymore.
It should be:
`const auto ret = std::ranges::remove(data, ' \r')`
Followed by
`data.erase(ret.begin(), ret.end());`
Please verify and apply changes if needed

diamante0018: Of course the variable names are at your discretion, I just typed an example of what I interpreted to he the correct usage from my phone :P

diamante0018: Alternatively also the compound statement (I believe that's what it is called), is available:
```
const auto [beg, end] = std::ranges::remove(data, ' \r');
data.erase(beg, end);
```

Brentdevent: Got it, I'll go with the last one seems the best option for readability. Thank you!

momo5502: decomposition seems best to me

momo5502 (approved):

diamante0018 (approved):
Oops

PR #248, opened 2023-02-27
feat: handle addtestclient when not host

Lierrmm:
fixes: #246


src/client/component/bots.cpp
```diff
@@ -126,6 +126,15 @@ namespace bots
...
 			command::add("spawnBot", [](const command::params& params)
 			{
+				if (!game::is_server())
+				{
+					const auto client_states = *reinterpret_cast<uint64_t*>(0x1576FB318_g);
+					if (!client_states)
+					{
+						return;
+					}
+				}
```
momo5502, line 136: please extract that into an is_host function

momo5502 (changes requested):

src/client/component/getinfo.cpp
```diff
@@ -78,6 +78,22 @@ namespace getinfo
...
+			return true;
+		}
+
+		const auto client_states = *reinterpret_cast<uint64_t*>(0x1576FB318_g);
+		if (!client_states)
+		{
+			return false;
+		}
+
+		return true;
```
momo5502, line 94:
```suggestion
		return client_states == 0;
```

momo5502 (changes requested):

src/client/component/getinfo.hpp
```diff
@@ -5,4 +5,5 @@ namespace getinfo
 	int get_max_client_count();
 	int get_client_count();
 	int get_bot_count();
+	bool is_host();
```
momo5502, line 8: i guess having that in there might make sense

Issue #253, opened 2023-02-27
Game crash when issuing command to Dedicated Server from Client


Trypticon8:
### Bug Type

Client

### BOIII Version

v0.0.1 (Latest)

### Environment (Operating System)

Microsoft Windows 11 (64-Bit)

### Reproducibility Rate

High: The bug can (almost) always be reproduced

### Steps to Reproduce

1. Upon joining any server, issue the "/spawnbot all" command. Crashes every single time for me

### Expected Result

I'd assume that as a client, you'd not be able to issue any commands to the server. I'd expect that nothing should happen, but instead my game crashes after issuing the command.

### Actual Result

After issuing the command, the game immediately freezes, crashes and opens the "BOIII ERROR" popup as shown below. Sidenote, this was tried on 2 different servers that were up at the time of me writing this
![image](https://user-images.githubusercontent.com/32724780/221706749-a4e52144-d01c-4665-9f5b-89ba1e45515c.png)

### (Optional) Proof and Additional Information

[minidumps.zip](https://github.com/momo5502/boiii/files/10845151/minidumps.zip)
https://user-images.githubusercontent.com/32724780/221711979-e5b66fda-9504-4c9c-a1a8-2da01180fe58.mp4

### Severity/Priority

Minor

Lierrmm:
linked: #245

PR #254, opened 2023-02-28
fix: if clientstate is 0 - we are not host

Lierrmm:
Should have been != rather than == on the client_state check.

resolves: #245 #253 and fixes #248


Lierrmm:
Seen your PR - so closing 😄

PR #279, opened 2023-03-04
Update addresses console component

Brentdevent:
Address updates to fix console component.

### DIFF
--- a/src/client/component/console.cpp
+++ b/src/client/component/console.cpp
@@ -54,7 +54,7 @@ namespace console
 			{
 				a.push(rbx);
 				a.mov(eax, 0x8030);
-				a.jmp(game::select(0x142333667, 0x140597527));
+				a.jmp(game::select(0x142332AA7, 0x140597527));
 			});
 
 			static_cast<void(*)(const char*)>(print_func)(message);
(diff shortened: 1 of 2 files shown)


momo5502 (approved):

momo5502:
How to we keep track of what got updated and what not?

src/client/component/console.cpp
```diff
@@ -251,7 +251,7 @@ namespace console
 			{
 				{
 					static utils::hook::detour sys_create_console_hook;
-					sys_create_console_hook.create(game::select(0x1423339C0, 0x140597880), sys_create_console_stub);
+					sys_create_console_hook.create(game::select(0x142332E00, 0x140597880), sys_create_console_stub);
```
momo5502, line 254: 0x1423333C0 is the correct address

Brentdevent: isn't Sys_ShowConsole at address 0x1423333C0?

Brentdevent:
> How to we keep track of what got updated and what not?

Good question, maybe we should do something like a checklist, taskboard?
Could also be useful to keep track of who is doing what so we're not colliding.

src/client/component/console.cpp
momo5502: jup, iwas dumb

momo5502:
> > How to we keep track of what got updated and what not?
> 
> Good question, maybe we should do something like a checklist, taskboard? Could also be useful to keep track of who is doing what so we're not colliding.

imo, the best thing to do is update all addresses in symbols.hpp, and then each component can be done individually

momo5502:
components for now are easy to distinguish with the REGISTER_COMPONENT_WORKING macro

Brentdevent:
Aah yes, right 👍🏼

Issue #306, opened 2023-03-08
Support usermaps on dedis


momo5502:
Last parameter of the connect function needs investigation

PR #308, opened 2023-03-08
Add usermap support

Brentdevent:
This PR allows clients to connect to dedicated servers/hosts that are running usermaps. 

Small side note -> switching from usermap back to a normal map causes a client/server index mismatch with zombie dedicated servers since I assume client does not unload the "usermap mod" fastfiles for some reason. I did not succeed to fix that issue so far. But It should not be a major issue atm. Client just needs to reconnect to server.

The other way around works just fine, same goes for switching between usermaps.

For now it also just throws a com_error when a user does not have an usermap installed.

This should solve #306

### DIFF
--- a/src/client/component/party.cpp
+++ b/src/client/component/party.cpp
@@ -1,16 +1,19 @@
 #include <std_include.hpp>
 #include "loader/component_loader.hpp"
 #include "game/game.hpp"
+#include "game/utils.hpp"
 
 #include "party.hpp"
 #include "network.hpp"
 #include "scheduler.hpp"
+#include "workshop.hpp"
 
 #include <utils/hook.hpp>
 #include <utils/string.hpp>
 #include <utils/info_string.hpp>
 #include <utils/cryptography.hpp>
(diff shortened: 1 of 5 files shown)


src/client/component/party.cpp
```diff
@@ -1,17 +1,20 @@
...
 #include "scheduler.hpp"
+#include "workshop.hpp"
 
 #include <utils/hook.hpp>
 #include <utils/string.hpp>
 #include <utils/info_string.hpp>
 #include <utils/cryptography.hpp>
 #include <utils/concurrency.hpp>
 
+
```
momo5502, line 17: why tho :P

momo5502 (approved):

src/client/component/party.cpp
Brentdevent: Ah lol, probably because I removed some includes after i was done debugging. I also see I forgot to remove  `#include "game/utils.hpp"`  It's no longer used.

I should have checked that more careful haha

PR #331, opened 2023-03-13
bdProfiles service

project-bo4:
fixes profile customization lockout(calling card and emblem only?)

Not tested on public games but clearly others players still wont see real info since your dw server is emulated locally and does not provide connectivity between players.

in case you want to adopt it with your system, basically one way might be somehow merging it with any of ddl's shared with dedicated servers; not so familiar with how it works.


4doorsmorehoes:
My man 🤝

momo5502 (approved):

PR #345, opened 2023-03-15
Disable change game mode in public lobby options.

joseupp:
This option is not necessary in public lobby since the game mode is set to playlist.


joseupp:
error

PR #344, opened 2023-03-15
Enable Dark matter camo.

joseupp:
Allows unlocking of the dark matter camo for the new stats menu

PR #349, opened 2023-03-16
More unlock all changes.

supitsmike:
- Unlock Dark Matter camo.
- Add unlock all specialists outfits.
- Change cg_unlockall_emblems_and_backings dvar to a more user friendly name. Add unlock specialists outfits to LUA menu.
- Remove stdcall from typedef.

### DIFF
--- a/data/ui_scripts/stats/__init__.lua
+++ b/data/ui_scripts/stats/__init__.lua
@@ -53,7 +53,7 @@ DataSources.MPStatsSettings = DataSourceHelpers.ListSetup( "MPStatsSettings", fu
 			value = 1
 		},
 	}, nil, updateDvar ))
-  table.insert( optionsTable, CoD.OptionsUtility.CreateDvarSettings( controller, "Unlock all Camos and Reticles", "All camos and reticles on weapons are unlocked.", "MPStatsSettings_unlockall_camos_and_reticles", "cg_unlockall_camos_and_reticles", {
+	table.insert( optionsTable, CoD.OptionsUtility.CreateDvarSettings( controller, "Unlock all Camos and Reticles", "All camos and reticles on weapons are unlocked.", "MPStatsSettings_unlockall_camos_and_reticles", "cg_unlockall_camos_and_reticles", {
 		{
 			option = "MENU_DISABLED",
 			value = 0,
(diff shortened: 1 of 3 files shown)


supitsmike:
I noticed when testing this that the dvars are not saved and set on launch anymore after setting them from the LUA menu.

Brentdevent:
> I noticed when testing this that the dvars are not saved and set on launch anymore after setting them from the LUA menu.

I can't replicate this issue, seems to work fine for me. Do the dvar values get written to the config file?

supitsmike:
> > I noticed when testing this that the dvars are not saved and set on launch anymore after setting them from the LUA menu.
> 
> I can't replicate this issue, seems to work fine for me. Do the dvar values get written to the config file?

The values are set in the config file but not getting set when the game loads.
image

I just cleared out those values from the config and relaunched it, after setting the values and closing the game they didn't get written to the config file.

Brentdevent:
> 

Writing to config file gets scheduled (10 seconds), are you closing the game perhaps before it takes place?
I was thinking initially my changes broke it but it shouldn't affect the writing to config file part.

Brentdevent:
> I just left the game open for 15 min after setting those dvars and then closed it but the config didn't update. Unless I'm doing something wrong I don't think its working.

Hmm, strange. If you change this pipline back to main does the issue go away?
https://github.com/momo5502/boiii/blob/5012affc9d0777155986a747b13b117b5c32de59/src/client/component/dvars.cpp#L197

supitsmike:
> > I just left the game open for 15 min after setting those dvars and then closed it but the config didn't update. Unless I'm doing something wrong I don't think its working.
> 
> Hmm, strange. If you change this pipeline back to main does the issue go away?
> 
> https://github.com/momo5502/boiii/blob/5012affc9d0777155986a747b13b117b5c32de59/src/client/component/dvars.cpp#L197

Yep, that fixes it. Do you want to investigate that so work with your setup or should I just commit that change on this as well.

Brentdevent:
> Yep, that fixes it. Do you want to investigate that so work with your setup or should I just commit that change on this as well.

Nah I'll have a look, I guess I'll have to find a different solution when to patch the dvar flags. I guess that's causing some issues right now. Config file needs to be read after the dvars have been patched otherwise you can't change them from config.

momo5502:
This does also unlock dark matter: #344 Which one is the right PR now? xD

supitsmike:
> This does also unlock dark matter: #344 Which one is the right PR now? xD

I feel like the way I unlocked dark matter is better because it’s just using a hook and not using the decompiled LUA like the other one

JariKCoding:
I agree with @supitsmike on this one, that Lua from the other PR looks nasty and I don't have the confidence in my decompiler that it decompiled such a complicated function correctly.

PR #357, opened 2023-03-18
Fix registering sessionmode bool dvars

Brentdevent:
This fixes the issue that the bool value does not get updated when registering sessionmode bool dvars.
I had to do some patches in the function `Dvar_SessionModeSetDefaultBool` on client side due arxan anti tampering. Calling this function would result in a freeze without these patches. I hope this won't cause any weird side effects. Atleast I did not notice them anymore once I properly patched it.

### DIFF
--- a/src/client/component/dvars_patches.cpp
+++ b/src/client/component/dvars_patches.cpp
@@ -66,12 +66,6 @@ namespace dvars_patches
 
 			// toggle ADS dof based on r_dof_enable
 			utils::hook::jump(0x141116EBB_g, utils::hook::assemble(dof_enabled_stub));
-
-			// De-Arxan Dvar_SessionModeSetDefaultBool
-			utils::hook::nop(0x1422D0EDB_g, 9);
-			utils::hook::jump(0x1422D0EE6_g, 0x1422D11BA_g);
-			utils::hook::jump(0x1422D11D1_g, 0x1422D11EC_g);
-			utils::hook::jump(0x1422D11F1_g, 0x1422D142C_g);
 		}
 	};
 }
--- a/src/client/component/dvars_patches.cpp
+++ b/src/client/component/dvars_patches.cpp
@@ -13,7 +13,7 @@ namespace dvars_patches
 	{
 		void patch_dvars()
 		{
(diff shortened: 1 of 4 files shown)


src/client/component/dvars_patches.cpp
```diff
@@ -66,6 +66,12 @@ namespace dvars_patches
 
 			// toggle ADS dof based on r_dof_enable
 			utils::hook::jump(0x141116EBB_g, utils::hook::assemble(dof_enabled_stub));
+
+			// De-Arxan Dvar_SessionModeSetDefaultBool
+			utils::hook::nop(0x1422D0EDB_g, 9);
+			utils::hook::jump(0x1422D0EE6_g, 0x1422D11BA_g);
+			utils::hook::jump(0x1422D11D1_g, 0x1422D11EC_g);
+			utils::hook::jump(0x1422D11F1_g, 0x1422D142C_g);
```
momo5502, line 74: those patches look wild. are you sure they work?

Brentdevent: Yea should be, I basically jump to execute these instructions which are in the arxan part.
image
This is part of assigning the values to DvarValue if I'm not mistaken which is later used.
After that I just skip the rest until the real part of the function starts again.

At first I forgot to do this, causing some weird issues because all these dvars using this function would get set incorrectly ofcourse. After adjusting it to this, the game seems to be functioning normal again with no freezes as far as I know.

However I'm still not 100% if it's okay to skip all that code. Since v18 seems to get altered by the code aswell or maybe it's just dead code. I'm not sure.

image

Maybe it would be safer/cleaner to create these DvarValues from scratch. But so far it seems to work without any issues.

momo5502: The arxan code seems simple to me, it checks that the return address is within the bo3 module. i will leave this pr open and try to write a generic patch for that.
while i trust that your patches work, i still don't want to risk any instabilities and want you to try faking the callstack instead. try introducing a fake stack frame which points to bo3

Brentdevent: Yea I fully understand, okay I'll give it a try in the next couple days :)

momo5502 (approved):

PR #360, opened 2023-03-18
Manually find UGC when SteamAPI fails to initialize

counter185:
On my Deck, BOIII always fails to initialize Steam which means it won't load workshop content. This means I have to run the game normally to play custom maps.
 
This PR fixes the problem by manually looking for subscribed items

I'm using two paths here, the relative one is needed in case the game is installed to a different drive.

Not sure if I should cache any values here because workshop content can be downloaded while the game is running


momo5502:
this is a workaround for an issue that should be inverstigated instead. why does steam initializatiom fail on your steam deck?
it works fine on mine

counter185:
no idea, wasted hours on trying to fix it, every version of proton there is including custom ones, nothing makes it want to actually connect to steam other than launching it unmodified. if it helps pinpoint what this could be caused by, i've noticed that any mod for any game that relies on dll injection will not work on both wine and proton, and that's not just on the deck, because i'm getting the same thing on my main pc with a linux dualboot.

momo5502:
well, are you using proton ge? help me reproduce it.figure out ehoch step fails, are you using the latest boiii version? cause i implemented steam deck support two days ago

counter185:
tried both proton ge and experimental and yes i'm on the very latest boiii commit. i also have to manually set it to a dev build or else the game will just say `Steam must be running to play this game!`. a normal release autoupdated build will run, but steam won't initialize (i can tell because the player name is `steamuser` instead of my steam name)

momo5502:
> tried both proton ge and experimental and yes i'm on the very latest boiii commit. i also have to manually set it to a dev build or else the game will just say `Steam must be running to play this game!`. a normal release autoupdated build will run, but steam won't initialize (i can tell because the player name is `steamuser` instead of my steam name)

username is not an indicator as it gets saved and has to be manually set. there are two methods in the steam_proxy: initialize and create_ugc. can you debug both and see where it fails to create the appropriate interfaces?

counter185:
neither of these 2 methods even get called

momo5502:
that's the root of the issue then and one needs to find out why

counter185:
the patched `SteamAPI_Init` doesn't get called either. pretty sure that's also what's happening with all the other xlabs mods since these too complain about steam not running.

momo5502:
> the patched `SteamAPI_Init` doesn't get called either. pretty sure that's also what's happening with all the other xlabs mods since these too complain about steam not running.

this is technically impossible unless you have fishy game files

counter185:
Okay just as I was recording a video to show what's going on, the mainline autoupdated builds started working properly, but running a debug build without `DEV_BUILD` defined still gives me a steam not running error, and a debug build with `DEV_BUILD` works but does not initialize steam.

Whether this PR gets merged or not i'll probably be keeping my own custom build with it because knowing linux this will happen again and at this point i just don't care about steamapi working as long as i can play custom maps

momo5502:
this pr will not get merged, as it is a workaround for something that is supposed to be working. i also do not recommend using custom builds, as you clearly do not comprehend the implications of dev_builds vs non dev builds. Main builds are built using a completely different configuration again.
The steam message is intended behaviour for a future ownership check. 
Why your steam api is not being initialized is the main issue which needs to be investigated.

momo5502:
Which is why i want you to close this pr and create an issue instead with a video recording which shows the bahviour with the official builds

PR #366, opened 2023-03-20
Change 4 character name limit to 3 characters

supitsmike:


src/client/component/network.cpp
```diff
@@ -291,6 +291,9 @@ namespace network
 
 			// Set cl_maxpackets to 100
 			utils::hook::set<uint8_t>(game::select(0x1412FF342, 0x140177A32), 100 - 15);
+
+			// Change 4 character name limit to 2 characters
```
momo5502, line 295: why network.cpp and not name.cpp or patches.cpp? also i suggest 3 chars to be the min amount

momo5502 (changes requested):

src/client/component/network.cpp
supitsmike: I did network.cpp and not name.cpp because name isn’t ran on the server I don’t think. But I can check the patches.cpp

I’ll change it to 3 chars

momo5502 (approved):

PR #393, opened 2023-03-25
feat(chat): g_deadChat functionality

diamante0018:


src/client/component/chat.cpp
```diff
@@ -1,19 +1,22 @@
 #include <std_include.hpp>
 #include "loader/component_loader.hpp"
-#include "chat.hpp"
```
momo5502, line 3: why?

momo5502 (changes requested):

src/client/component/chat.cpp
diamante0018: It is grouped together with the other component's header at line 10.

PR #420, opened 2023-04-01
feat(chat): g_deadChat functionality

diamante0018:


diamante0018:
Dss0 and I tested this and it works as intended on Search and Destroy game mode.
As you left a review on my old PR @momo5502 I would like to address it by saying I personally don't prefer how you put chat.hpp on top of the header file. but if that's a deal breaker for you let me know.

momo5502:
> Dss0 and I tested this and it works as intended on Search and Destroy game mode.
> As you left a review on my old PR @momo5502 I would like to address it by saying I personally don't prefer how you put chat.hpp on top of the header file. but if that's a deal breaker for you let me know.

there is a reason they are at the top. because headers should compile on their own (ignoring pch). putting them lower im the cpp file will implicitly deckare symbols from headers higher in the order which might make them compile accidentally, even though the header is missing an include or a forwad declaration.

diamante0018:
> > Dss0 and I tested this and it works as intended on Search and Destroy game mode.
> > As you left a review on my old PR @momo5502 I would like to address it by saying I personally don't prefer how you put chat.hpp on top of the header file. but if that's a deal breaker for you let me know.
> 
> there is a reason they are at the top. because headers should compile on their own (ignoring pch). putting them lower im the cpp file will implicitly declare symbols from headers higher in the order which might make them compile accidentally, even though the header is missing an include or a forwad declaration.

Updated the pr restoring old header order

momo5502 (approved):

Issue #365, opened 2023-03-19
Usermap refresh command?


StickyStickx:
### Duplicate Issue Check

- [X] I have searched the list of issues on this repository. I confirm that this is not a duplicate of any open or closed issue(s).

### Your feature request or suggestion:

So I was streaming some BO3 last night and I thought it would be a good idea to use the boiii client instead of steam just in case there are still some vulnerabilities. I was installing custom maps through steam, but they weren't being added to my map list on the boiii client unless I restarted it first. I was wondering if its possible to refresh the custom maps / workshop folder by using a command instead of rebooting the game every time?

PR #441, opened 2023-04-04
refresh custom maps list when entering mp/zm menu

WantedDV:
Going into the Multiplayer/Zombies menu now refreshes the custom maps list so you don't have to reopen the game.

Closes https://github.com/momo5502/boiii/issues/365


momo5502:
putting that in stats lua file seems off to me

momo5502 (approved):

PR #445, opened 2023-04-04
fix: restart button is never shown on online sessions

SergiX44:
- The restart button is never show to the host in private zombie sessions


diamante0018 (approved):

diamante0018:
This script appears to have been messed up by the decompiler that was used on the original compiled game script.
Nice catch. I'm gonna approve the workflow run so momo can later merge this.

PR #456, opened 2023-04-07
Distribute profile infos to connected clients

momo5502:


momo5502:
PR is unfinished, but causes a lot of merge conflicts with contributions. So the auth protocol is disabled and can be enabled by reverting 6d0d3510646121acf41f3164e283b60201fa2d81

src/client/game/symbols.hpp
```diff
@@ -144,7 +144,7 @@ namespace game
 	WEAK symbol<void(int localClientNum)> CL_CheckKeepDrawingConnectScreen{0x1413CCAE0};
 
 	// Scr
-	WEAK symbol<void(scriptInstance_t inst, int value)> Scr_AddInt{0x1412E9870, 0x14016F160};
+	WEAK symbol<void(scriptInstance_t inst, int value)> Scr_AddInt{0x0, 0x14016F160};
```
Brentdevent, line 147: Merge mistake I assume?

src/client/game/structs.hpp
```diff
@@ -1544,14 +1544,31 @@ namespace game
...
+		uint64_t xuid;
+		char __pad2[0xB5D84];
+		int guid;
+		char __pad3[0x8];
 		bool bIsTestClient;
-		char __pad2[0x29DAC];
+		char __pad4[0x29DAC];
+	};
+
+	static_assert(sizeof(client_s) <= 0xE5110);
```
Brentdevent, line 1559: Shouldn't this be "==" instead of "<=" or is their reason for it to be "<="

src/client/game/symbols.hpp
momo5502: indeed, thx

src/client/game/structs.hpp
momo5502: that's fixed on the branch, but the size is still correct, only the comparison is wrong

src/client/game/symbols.hpp
momo5502: fixed in https://github.com/momo5502/boiii/commit/994cfb5af90361955223677702f66d9dc07acd9d

PR #483, opened 2023-04-10
cleanup game settings

diamante0018:


diamante0018:
the const std::string return value is quite undesirable as it prevents optimizations.

src/client/component/game_settings.cpp
```diff
@@ -11,54 +11,54 @@ namespace gamesettings
 	namespace
 	{
 		// <name, path>
-		std::unordered_map<std::string, std::string> gamesettings_files;
+		std::unordered_map<std::string, std::string> game_settings_files;
 
-		const std::string get_gamesettings_name(const std::vector<std::string>& sub_strings)
```
momo5502, line 16: good. i think the const prevented copy elision.

src/client/component/game_settings.cpp
```diff
@@ -11,54 +11,54 @@ namespace gamesettings
...
+		std::string get_game_settings_name(const std::vector<std::string>& sub_strings)
 		{
 			if (sub_strings.size() > 2)
 			{
 				return sub_strings[sub_strings.size() - 2] + '/' + sub_strings[sub_strings.size() - 1];
 			}
 
-			return "";
+			return std::string();
```
momo5502, line 23: {}

src/client/component/game_settings.cpp
```diff
@@ -11,54 +11,54 @@ namespace gamesettings
...
 		}
 
-		const std::string get_gamesettings_path(const std::string& name)
+		std::string get_game_settings_path(const std::string& name)
 		{
-			const auto itr = gamesettings_files.find(name);
-			return (itr == gamesettings_files.end()) ? "" : itr->second;
+			const auto itr = game_settings_files.find(name);
+			return (itr == game_settings_files.end()) ? std::string() : itr->second;
```
momo5502, line 29: {} instead of std::string()

momo5502 (changes requested):
good PR. while the PR yesterday was also very good, i was in a rush yesterday and didn't properly review it, to be able to spot those things. thanks for cleaning it up.

src/client/component/game_settings.cpp
diamante0018: This will cause a compilation error because this syntax is not valid in this context. Due to the ternary operator that is.
I propose std::string{}

```
1>game_settings.cpp(29,48): error C2059: syntax error: '{'
1>game_settings.cpp(29,48): error C2143: syntax error: missing ';' before '{'
1>game_settings.cpp(29,51): error C2059: syntax error: ':'
```

momo5502: nvm, clang also cannot. then fuck it.

momo5502 (approved):

src/client/component/game_settings.cpp
diamante0018: No, it does not work, unfortunately.

PR #494, opened 2023-04-10
fix: support for hardcore on the server browser

JariKCoding:
Added support for the hardcore skull icon in the server browser
image

### DIFF
--- a/src/client/component/getinfo.cpp
+++ b/src/client/component/getinfo.cpp
@@ -113,7 +113,7 @@ namespace getinfo
 				info.set("gamemode", utils::string::va("%i", Com_SessionMode_GetGameMode()));
 				info.set("sv_running", utils::string::va("%i", game::is_server_running()));
 				info.set("dedicated", utils::string::va("%i", game::is_server() ? 1 : 0));
-				info.set("hc", utils::string::va("%i", game::Com_GametypeSettings_GetUInt("hardcoremode", false)));
+				info.set("hc", utils::string::va("%u", game::Com_GametypeSettings_GetUInt("hardcoremode", false)));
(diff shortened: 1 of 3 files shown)


src/client/component/getinfo.cpp
```diff
@@ -112,7 +112,8 @@ namespace getinfo
 				info.set("playmode", utils::string::va("%i", game::Com_SessionMode_GetMode()));
 				info.set("gamemode", utils::string::va("%i", Com_SessionMode_GetGameMode()));
 				info.set("sv_running", utils::string::va("%i", game::is_server_running()));
-				info.set("dedicated", utils::string::va("%i", game::is_server() ? 1 : 0));
+				info.set("dedicated", utils::string::va("%i", game::is_server() ? 1 : 0));
+				info.set("hc", utils::string::va("%i", game::Com_GametypeSettings_GetUInt("hardcoremode", false)));
```
diamante0018, line 116: now that i think of it, why is utils::string::va being used this much in this file when std::to_string exists? 😂 
probably a question for another day, patch looks good.

diamante0018 (changes requested):

diamante0018 (approved):

src/client/component/getinfo.cpp
momo5502: that really screams like good old c&p 😂

diamante0018:
@momo5502 could u merge this so i can then make a new pr for ::va?

momo5502:
> @momo5502 could u merge this so i can then make a new pr for ::va?
> Or you wanted myself or jarik do address it right away?

you uave to close open comments i think. cant see it on my phone

PR #505, opened 2023-04-11
Support packet fragmentation

momo5502:


diamante0018 (approved):
Superb! This is much better than simply "sending bigger packets"

momo5502:
> Superb! This is much better than simply "sending bigger packets"

maybe simpler is the wrong word. took way more effort than patching a few sizes 😂 
but this is definitely way more robust

Brentdevent (approved):
Don't see anything unusual, don't understand it all but I think it should be good to go 😄

Issue #415, opened 2023-03-30
Hide Custom Maps button if no usermaps installed


WantedDV:
### Bug Type

Client

### BOIII Version

v0.0.1 (Latest)

### Environment (Operating System)

Microsoft Windows 10 (64-Bit)

### Reproducibility Rate

High: The bug can (almost) always be reproduced

### Steps to Reproduce

1. launch MP
2.  goto map selection menu and click on custom maps

### Expected Result

Hide Custom maps button if there are no mp maps in `usermaps` or show a general preview image.

### Actual Result

![custommap](https://user-images.githubusercontent.com/122710241/228987856-5d2bccbd-5a0a-45df-9d8e-ffb5d90ef642.PNG)

### (Optional) Proof and Additional Information

This is not an issue if custom maps are installed.

![map2](https://user-images.githubusercontent.com/122710241/228987911-b89f5e92-1890-4eb9-8ac8-7820a8fb50b0.PNG)

### Severity/Priority

Minor

momo5502:
pretty useless to spend time on this imo

Brentdevent:
For me the button disappears, I can't replicate this. Got an empty usermaps folder in the root directory + steam is closed.
Anyway I agree with momo. 
![image](https://user-images.githubusercontent.com/70229620/229276911-b58961ea-6cde-4058-a707-049058e266b5.png)

FragsAreUs:
this issue is related to people who dont own the game threw steam as bo3 is/was intented to use the steam workshop to download custom maps and what not. So imo it is also a usless thing to tackle because I am pretty sure if you did the same thing if you own it on steam it would display the same issue.

mjkzy:
this should be marked as low priority and good first issue as it's cool for a beginner to tackle something like this :D

Issue #322, opened 2023-03-11
Barracks not available while mods are loaded


Mikeehh:
### Bug Type

Client

### BOIII Version

v0.0.1 (Latest)

### Environment (Operating System)

Microsoft Windows 11 (64-Bit)

### Reproducibility Rate

High: The bug can (almost) always be reproduced

### Steps to Reproduce

1. Load a mod
2. Try to view your Barracks in the Menu

### Expected Result

Barracks should be available

### Actual Result

Barracks are not available

### (Optional) Proof and Additional Information

Would be nice to access the Barracks when mods are loaded since you can rank up while mods are loaded now, mostly just to be able to prestige and such.

A few more minor issues while mods are loaded:
- Double XP icons do not show up in the pause menu, even if it is enabled.
- You cannot view your daily challenge, shows up as blank in the Featured section (even though you can still complete it)

### Severity/Priority

Minor

PR #508, opened 2023-04-11
Frontend menus

WantedDV:
### Improvements
- Cleaned up the menu buttons and spacings.
- Added a large server button to the multiplayer menu
- Removed any indication that the lobby is "searching" for other players.
- Find Game is renamed to "Combat Training" to avoid confusion.

### Fixes
- Disable custom map category button if no **mp** usermaps installed. #415
- Barracks is no longer grayed out when a mod is loaded. #322
- Mapvote in the public lobby now works correctly to set the map and doesn't kick them out of the menu.

Here is a snapshot of the current menu
 m2 m1

Here are some of the things that were removed.
m3
m4
m5


JariKCoding:
We need to find a way to change data sources without completely overriding them. I'll help @WantedDV searching for this.

momo5502:
> We need to find a way to change data sources without completely overriding them. I'll help @WantedDV searching for this.

is this ready? from the description it looks really good. but the menus are hard to judge for me.

JariKCoding (approved):

momo5502 (approved):

Issue #488, opened 2023-04-10
Mods don't work on Dedicated Servers


Dss0:
### Bug Type

Dedicated Server

### BOIII Version

v0.0.1 (Latest)

### Environment (Operating System)

Microsoft Windows 10 (64-Bit)

### Reproducibility Rate

High: The bug can (almost) always be reproduced

### Steps to Reproduce

1. Set fs_game to a mod in the startup parameters, this can either be a folder in /identities/dedicatedpc/mods or a workshop id
2. Start the server

### Expected Result

The server loads the mod

### Actual Result

The server does change fs_game (if the folder exists) but does not load any of the mod's fastfiles.

### Severity/Priority

Medium

PR #513, opened 2023-04-12
Mods support for dedicated servers

Brentdevent:
- Client auto loads the same mod as the server is running, upon connection
- Mod name is read from the workshop.json and send to client to visualize the mod name in the serverbrowser

**Important note**: The mod name varies in size since `m_szGameTags` has a max buffer size of 128 characters. So because of the gametype and the other parameters it causes the mod name to be shortened to the available size that is left. The buffer is just way to small, it seems kinda tedious to increase this. Atleast I don't see an easy way.

This also seems to occur on the steam version, so I guess it should be fine for now.
 
Dedicated servers were already able to load mods by using the following path: `Call of Duty Black Ops III\UnrankedServer\mods\<publisher_id>\zone`

Closes issue https://github.com/momo5502/boiii/issues/488

Some files contain wrong line endings, I did not change it. To make this PR more easier to read.
And I assume this belongs to a different PR.

### DIFF
--- a/src/client/component/workshop.cpp
+++ b/src/client/component/workshop.cpp
@@ -88,12 +88,18 @@ namespace workshop
 
 		if (json_str.empty())
 		{
-			printf("[ Workshop ] Workshop.json has not been found in mod folder: %s", mod_id.data());
+			printf("[ Workshop ] workshop.json has not been found in mod folder: %s\n", mod_id.data());
 			return mod_id;
 		}
 
 		rapidjson::Document doc;
-		doc.Parse(json_str);
+		const rapidjson::ParseResult parse_result = doc.Parse(json_str);
+
+		if (parse_result.IsError() || !doc.IsObject())
+		{
+			printf("[ Workshop ] Unable to parse workshop.json\n");
+			return mod_id;
+		}
 
 		if (doc.HasMember("Title"))
 		{
@@ -108,7 +114,7 @@ namespace workshop
 		}
 		else
 		{
-			printf("[ Workshop ] Workshop.json has no \"Title\" member.");
+			printf("[ Workshop ] workshop.json has no \"Title\" member.\n");
 			return mod_id;
 		}
 	}
--- a/src/client/component/workshop.cpp
+++ b/src/client/component/workshop.cpp
@@ -63,7 +63,7 @@ namespace workshop
 		}
 	}
 
-	const std::string get_mod_name(const std::string& mod_id)
+	std::string get_mod_name(const std::string& mod_id)
 	{
 		if (mod_id == "usermaps")
 		{
@@ -113,7 +113,7 @@ namespace workshop
 		}
 	}
 
-	const std::string get_usermap_publisher_id(const std::string& mapname)
+	std::string get_usermap_publisher_id(const std::string& mapname)
 	{
 		const auto total_usermaps = *reinterpret_cast<unsigned int*>(0x1567B3580_g);
 
--- a/src/client/component/workshop.cpp
+++ b/src/client/component/workshop.cpp
@@ -65,26 +65,15 @@ namespace workshop
 
 	std::string get_mod_name(const std::string& mod_id)
 	{
-		if (mod_id == "usermaps")
+		if (mod_id == "usermaps" || !game::is_server())
 		{
 			return mod_id;
 		}
 
 		const utils::nt::library host{};
-		std::string path;
-
-		if (game::is_server())
-		{
-			const auto base_path = host.get_folder().generic_string();
-			path = utils::string::va("%s/mods/%s/zone/workshop.json", base_path.data(), mod_id.data());
-		}
-		else
-		{
-			return mod_id;
-		}
-
-		std::string json_str;
-		utils::io::read_file(path, &json_str);
+		const auto base_path = host.get_folder().generic_string();
+		const auto path = utils::string::va("%s/mods/%s/zone/workshop.json", base_path.data(), mod_id.data());
+		const auto json_str = utils::io::read_file(path);
 
 		if (json_str.empty())
 		{
@@ -112,11 +101,9 @@ namespace workshop
 
 			return title;
 		}
-		else
-		{
-			printf("[ Workshop ] workshop.json has no \"Title\" member.\n");
-			return mod_id;
-		}
+
+		printf("[ Workshop ] workshop.json has no \"Title\" member.\n");
+		return mod_id;
 	}
 
 	std::string get_usermap_publisher_id(const std::string& mapname)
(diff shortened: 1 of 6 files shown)


src/client/component/workshop.cpp
```diff
@@ -72,6 +63,107 @@ namespace workshop
 		}
 	}
 
+	const std::string get_mod_name(const std::string& mod_id)
```
diamante0018, line 66: remove const, it prevents optimizations.

diamante0018 (changes requested):

src/client/component/workshop.cpp
```diff
@@ -72,6 +63,107 @@ namespace workshop
...
+			return title;
+		}
+		else
+		{
+			printf("[ Workshop ] Workshop.json has no \"Title\" member.");
+			return mod_id;
+		}
+	}
+
+	const std::string get_usermap_publisher_id(const std::string& mapname)
```
diamante0018, line 116: Same here as well.
And source: https://clang.llvm.org/extra/clang-tidy/checks/readability/const-return-type.html

Brentdevent: Oh interesting learned something new, thank you!

src/client/component/workshop.cpp
```diff
@@ -72,6 +63,107 @@ namespace workshop
...
+		utils::io::read_file(path, &json_str);
+
+		if (json_str.empty())
+		{
+			printf("[ Workshop ] Workshop.json has not been found in mod folder: %s", mod_id.data());
+			return mod_id;
+		}
+
+		rapidjson::Document doc;
+		doc.Parse(json_str);
```
diamante0018, line 96: I strongly advise to catch parse errors like this: https://github.com/XLabsProject/master/blob/master/src/patreon.cpp#L23

diamante0018 (changes requested):

momo5502 (approved):

src/client/component/workshop.cpp
```diff
@@ -72,6 +63,113 @@ namespace workshop
...
+			const auto base_path = host.get_folder().generic_string();
+			path = utils::string::va("%s/mods/%s/zone/workshop.json", base_path.data(), mod_id.data());
+		}
+		else
+		{
+			return mod_id;
+		}
+
+		std::string json_str;
+		utils::io::read_file(path, &json_str);
```
momo5502, line 87: there is a shorter read_file that directly returns a string
```suggestion
		std::string json_str = utils::io::read_file(path);
```

src/client/component/workshop.cpp
```diff
@@ -72,6 +63,113 @@ namespace workshop
...
+
+		if (game::is_server())
+		{
+			const auto base_path = host.get_folder().generic_string();
+			path = utils::string::va("%s/mods/%s/zone/workshop.json", base_path.data(), mod_id.data());
+		}
+		else
+		{
+			return mod_id;
+		}
```
momo5502, line 84: why no early return here?

```suggestion
		if (!game::is_server())
		{
			return mod_id;
		}
		
		const utils::nt::library host{};
		const auto base_path = host.get_folder().generic_string();
		std::string path = utils::string::va("%s/mods/%s/zone/workshop.json", base_path.data(), mod_id.data());
```

src/client/component/workshop.cpp
momo5502: or you respect the return value of your current read_file call. it returns a bool whether reading was successful or not

src/client/component/workshop.cpp
```diff
@@ -72,6 +63,113 @@ namespace workshop
...
+				title.resize(31);
+			}
+
+			return title;
+		}
+		else
+		{
+			printf("[ Workshop ] workshop.json has no \"Title\" member.\n");
+			return mod_id;
+		}
```
momo5502, line 119: you can remove the else scope here, as the if has a return statement
```suggestion

		printf("[ Workshop ] workshop.json has no \"Title\" member.\n");
		return mod_id;
```

src/client/component/workshop.cpp
```diff
@@ -72,6 +63,113 @@ namespace workshop
...
+	}
+
+	std::string get_usermap_publisher_id(const std::string& mapname)
+	{
+		const auto total_usermaps = *reinterpret_cast<unsigned int*>(0x1567B3580_g);
+
+		for (unsigned int i = 0; i < total_usermaps; ++i)
+		{
+			const auto usermap_data = reinterpret_cast<game::workshop_data*>(0x1567B3588_g + (sizeof(game::workshop_data) * i));
```
momo5502, line 128: this code only runs either on client or on dedi, not sure. wouldn't it be desirable to have it work for both?

src/client/component/workshop.cpp
```diff
@@ -72,6 +63,113 @@ namespace workshop
...
+	bool check_valid_mod_id(const std::string& mod)
+	{
+		if (mod.empty() || mod == "usermaps")
+		{
+			return true;
+		}
+
+		if (!has_mod(mod))
+		{
+			game::Com_Error(0, "Can't find mod with publisher id: %s!\nMake sure you're subscribed to the workshop item.", mod.data());
```
momo5502, line 158: Com_Error is super evil. please don't use it. it prevents object destruction and thus casues memory leaks as it unwinds the stack in an unclean way, at least legacy setjmp implementations don't do proper unwinding.

momo5502 (changes requested):

src/client/component/workshop.cpp
Brentdevent: It runs only on the client yea, due we're using a non steam method for dedicated servers to load user content. This data section doesn't contain the publisherId.

But I have been thinking about using the publisherId as folder name for usermaps too. But this most likely require some more changes so perhaps this something for another PR.

Brentdevent:
> > But I have been thinking about using the publisherId as folder name for usermaps too. But this most likely require some more changes so perhaps this something for another PR.
> 
> If using publisherId as folder name is the best way to go then we should probably update the ui so it still shows mod names instead of this: image Similar for usermaps if you make that change. This is probably something I can do.
> 
> The only other downside is that content in these folders will be not easily identifiable unless you open up the workshop file in each folder to see what it is. At the same time though, this is how steam does it.

Well I'm still not certain about a few things, should we support add mod/usermaps support for non-steam copies to play online? 

Maybe a better alternative would be to iterate through the steam workshop folder and add the content ourselves to the usermap and mods data pool whenever steam is not detected.

But this will be pointless to implement at the moment if there will be steam check later.

Also would it be desirable for server owners to install steam on their VPS to download workshop content? Or should we stick with the current method we're using to load usercontent.

I have to admit the current implementation of the mod support is not ideal for non-steam copies indeed. Honestly I haven't even tested it. So I'm not even certain if it'll work for them.

momo5502 (approved):

momo5502:
Anything else? @WantedDV @Brentdevent ? Do you think changes are still needed?

Brentdevent:
Should be fine for now, I'll look into a way to improve it for offline use since I understand the concerns. Not entirely sure yet what route I'll be taking. But I'll improve it in a later PR. I have some ideas but I'll have to see which one will be a better one.

It makes things a lot more complicated, to keep it working in a clean way but to satisfy both ways of loading user content.

diamante0018 (approved):

PR #519, opened 2023-04-12
feat: add support for favorites in the server browser

JariKCoding:
Writes favorited servers to players/user/favorite_servers.csv

image


src/client/component/server_list.cpp
```diff
@@ -75,17 +78,63 @@ namespace server_list
...
-			if (state)
+		void write_favorite_servers()
+		{
+			std::string servers_buffer = "";
+			for (auto itr : favorite_servers)
 			{
-				auto botCount = atoi(game::Info_ValueForKey(serverInfo.tags, "bots"));
-				game::Lua_SetTableInt("botCount", botCount, state);
-			}
+				servers_buffer.append(utils::string::va("%u,%u\n", itr.addr, itr.port));
```
momo5502, line 104: why use a csv and not just directly store 127.0.0.1:27017 in every line
then you can just split the lines and directly pass the line to network::address_from_string or whateever it's called

momo5502, line 104: why use a csv and not just directly store 127.0.0.1:27017 in every line
then you can just split the lines and directly pass the line to network::address_from_string or whateever it's called

momo5502 (changes requested):

src/client/component/server_list.cpp
```diff
@@ -145,9 +217,14 @@ namespace server_list
...
-			}, scheduler::async, 200ms);
+			}, scheduler::async, 200ms);
+
+			lua_serverinfo_to_table_hook.create(0x141F1FD10_g, lua_serverinfo_to_table_stub);
 
-			lua_serverinfo_to_table_hook.create(0x141F1FD10_g, lua_serverinfo_to_table_stub);
+			scheduler::loop([]
+			{
+				read_favorite_servers();
+			}, scheduler::main);
```
momo5502, line 227: loop? are you sure you want an infinite loop that fires every frame?

src/client/component/server_list.cpp
```diff
@@ -112,6 +153,37 @@ namespace server_list
...
+			{
+				favorite_servers.erase(it);
+				break;
+			}
+		}
+
+		write_favorite_servers();
+	}
+
+	bool has_favorited_server(game::netadr_t addr)
```
momo5502, line 183: const game::netadr_t&

src/client/component/server_list.cpp
```diff
@@ -25,7 +26,9 @@ namespace server_list
 			callback callback{};
 		};
 
-		utils::concurrency::container<state> master_state;
+		utils::concurrency::container<state> master_state;
+
+		std::vector<game::netadr_t> favorite_servers{};
```
momo5502, line 31: i recommend using an unordered_set instead of a vector, that implicitly gets rid of the need of all duplication checks

src/client/steam/interfaces/matchmaking_servers.cpp
```diff
@@ -198,7 +199,54 @@ namespace steam
...
+					{
+						srvs = {};
+						srvs.reserve(s.size());
+
+						for (auto& address : s)
+						{
+							if (!server_list::has_favorited_server(address))
+							{
+								continue;
+							}
```
momo5502, line 234: so you query all servers and filter out all but the favorites. interesting. should be fine for now, but on the long run i recommend querying only the favorites

JariKCoding: I've changed it so it only queries your favorites. I've also made some big changes to keep the lists seperate so when switching between tabs fast internet servers don't end up on the favorites tab.

src/client/component/server_list.cpp
```diff
@@ -180,11 +175,10 @@ namespace server_list
 		write_favorite_servers();
 	}
 
-	bool has_favorited_server(game::netadr_t addr)
+	std::unordered_set<game::netadr_t>* get_favorite_servers()
```
momo5502, line 178: std::unordered_set<game::netadr_t>&

no need to return a pointer here, just return a ref

momo5502 (changes requested):

src/client/component/server_list.cpp
```diff
@@ -112,7 +153,32 @@ namespace server_list
...
+			}
+		}
+
+		write_favorite_servers();
+	}
+
+	std::unordered_set<game::netadr_t>& get_favorite_servers()
+	{
+		return favorite_servers;
+	}
```
momo5502, line 181: i have one last question here, are you sure this set is not used by multiple threads? from looking at the code i think it's safe, but can you maybe cofirm that? maybe even verify that? or should we wrap it in a concurrency container? just to be safe to avoid any races?

momo5502 (approved):

PR #558, opened 2023-04-18
Improve mod/usermap support

Brentdevent:
- Better support for offline usermap/mods loading, You are now able to just copy & paste the workshop folder in the usermaps/mods folder. The additional zone folder is no longer used.
- workshop.json data gets written to usermaps/mods pool and is required to make it work correctly to play online.  Without it should still work but folder names need to be synced could still be useful for mod developers. But it will print out a warning to console since the publisherId will not be numerical in this case.
- Mods/usermaps which are loaded offline now show up in the menus with the correct data just like the ones from steam workshop.
- This allows dedicated servers to rename their folders to whatever they want, although it seems this causes issues with some mods for unknown reasons. For example I got a clientField mismatch with the mod called mwz_lite after renaming the folder from publisherID to mwz_lite. I assume it has something to do with mod itself since with another mod this doesn't occur.


momo5502 (commented):
looks good at first sight, but will do an in-depth review tomorrow

src/client/component/party.cpp
```diff
@@ -181,15 +181,15 @@ namespace party
...
+			const auto mod_id = info.get("modId");
 
 			//const auto hostname = info.get("sv_hostname");
 			const auto playmode = info.get("playmode");
 			const auto mode = static_cast<game::eModes>(std::atoi(playmode.data()));
 			//const auto xuid = strtoull(info.get("xuid").data(), nullptr, 16);
 
 			scheduler::once([=]
-			{
+			{
```
momo5502, line 192: why? xD

src/client/component/workshop.cpp
```diff
@@ -30,92 +29,182 @@ namespace workshop
...
-			return utils::hook::invoke<bool>(0x1420D6380_g, type, name, a3);
+			utils::string::copy(item.title, doc["Title"].GetString());
+			utils::string::copy(item.description, doc["Description"].GetString());
+			utils::string::copy(item.folderName, doc["FolderName"].GetString());
+			utils::string::copy(item.publisherId, doc["PublisherID"].GetString());
+			item.publisherIdInteger = std::stoul(item.publisherId);
 		}
 
-		game::workshop_data* load_usermap_stub(const char* mapname)
+		void load_usermap_content_stub(void* usermapsCount, int type)
```
momo5502, line 88: where snake case bro? usermapsCount?

src/client/component/workshop.cpp
```diff
@@ -30,92 +29,182 @@ namespace workshop
...
+				auto& usermap_data = game::usermapsPool[i];
 
-			return utils::hook::invoke<game::workshop_data*>(0x1420D5700_g, name);
+				// foldername == title -> non-steam workshop usercontent
+				if (std::strcmp(usermap_data.folderName, usermap_data.title))
+				{
+					continue;
+				}
+
+				load_workshop_data(usermap_data);
```
momo5502, line 102: am i wrong or don't you need a break afterwards?

src/client/component/workshop.cpp
```diff
@@ -30,92 +29,182 @@ namespace workshop
...
+			for (unsigned int i = 0; i < *game::modsCount; ++i)
+			{
+				auto& mod_data = game::modsPool[i];
+
+				if (std::strcmp(mod_data.folderName, mod_data.title))
+				{
+					continue;
+				}
+
+				load_workshop_data(mod_data);
```
momo5502, line 119: break?

src/common/utils/string.cpp
```diff
@@ -65,6 +65,15 @@ namespace utils::string
...
 	}
 
+	bool is_numeric(const std::string& text)
+	{
+		if (text.size() <= 0 || !std::isdigit(text[0])) return false;
+		return std::ranges::all_of(text.begin(), text.end(), [](const char input)
+		{
+			return std::isdigit(input) != 0;
+		});
+	}
```
momo5502, line 75: while your code is correct, i always used std::to_string(atoi(text.data())) == text
has a heap allocation (probably not with sso) which yours doesn't but it's a super concise one liner. but yours is probably better. just as a hint

momo5502 (changes requested):
Honestly impressed by the code you write. It gets cleaner and cleaner with every PR. Very cool :)
Still a few remarks 😂 But only minor things

src/client/component/workshop.cpp
Brentdevent: Nah, I'm iterating over every usermap/mod, check if it's loaded from the usermaps/mods folder if so I load the data from the corresponding workshop.json.

src/client/component/workshop.cpp
Brentdevent: See the other reply.

src/client/component/workshop.cpp
Brentdevent: Good catch, sorry oversight, tend to make this mistake by switching between IDA and vs haha

src/client/component/party.cpp
Brentdevent: Performance improvements 😎 nah kidding, missed this change.

src/common/utils/string.cpp
Brentdevent: Yea I did the same at first, untill I noticed that the publisherID's are bigger than the max size of an integer. So I went with this implementation. I was not aware of the existence of stoul at the time. If you want I can change it. It's whatever you prefer.

momo5502:
Has conflicts. If you resolve them, we can merge it :)

momo5502 (approved):

Brentdevent:
I changed one more thing, it was possible for clients to load a mod and join a non-modded server, so I do an additional check and unload the mod if this is the case.

src/client/component/workshop.cpp
```diff
@@ -30,92 +29,182 @@ namespace workshop
...
+			{
+				printf("[ Workshop ] workshop.json is invalid:\n%s\n", path);
+				return;
+			}
+
+			utils::string::copy(item.title, doc["Title"].GetString());
+			utils::string::copy(item.description, doc["Description"].GetString());
+			utils::string::copy(item.folderName, doc["FolderName"].GetString());
+			utils::string::copy(item.publisherId, doc["PublisherID"].GetString());
+			item.publisherIdInteger = std::stoul(item.publisherId);
```
diamante0018, line 85: Potentially unsafe as this function will throw if it can't parse the input and cause a crash

src/common/utils/string.cpp
```diff
@@ -66,6 +66,15 @@ namespace utils::string
 		return std::equal(substring.rbegin(), substring.rend(), text.rbegin());
 	}
 
+	bool is_numeric(const std::string& text)
+	{
```
diamante0018, line 70: Type is size_t so <= 0 is not correct as it can't never be less than 0. Check for empty

PR #590, opened 2023-04-24
small fixes

diamante0018:
Readability of std::strcmp is important in my opinion, that is why extensions like re-sharper & clang-tidy recommend to always explicitly comparing the result.

Other than that is_numeric needs the other fixes I already did quite some time ago to string.cpp where the character should be converted to unsigned char to avoid UB. Additionally, the readability of the function was improved.

FInally fixed a would be crash caused by stoul being a throwable function.


momo5502 (approved):
Good PR

PR #592, opened 2023-04-24
Make usercontent zone folder toggleable

Brentdevent:
It seems mod developers prefer to keep the zone folder which I can understand. So I came up with this idea. 
`enable_zone_folder 1` loads the mods/usermaps fastfiles from the zone folder just like how it was before.

I also added a command `userContentReload` that reloads the usermaps/mods in their pools.

Is there a way to reduce the redundancy of the asm stubs? Because they are very similiar only the strings + addresses differ.
I feel like this might not be very clean but I don't see a way to do this differently. So let me know if it needs some changes :)

### DIFF
--- a/src/client/component/workshop.cpp
+++ b/src/client/component/workshop.cpp
@@ -3,6 +3,7 @@
 #include "workshop.hpp"
 
 #include "game/game.hpp"
+#include "game/utils.hpp"
 
 #include <utils/hook.hpp>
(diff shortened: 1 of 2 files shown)


momo5502 (approved):
There are ways to reduce redundancy, but I don't think it makes sense, because right now everything is perfectly readable. I would not want to sacrifice that

PR #595, opened 2023-04-25
rcon: add flooding protection

diamante0018:
Original review: https://github.com/XLabsProject/iw4x-client/pull/654
Although please let me know if the code can be improved.
I also took the liberty of cleanup the format of some other files.

A steam file will have a huge diff, that is because I changed the line endings to our favorite CRLF


src/client/component/rcon.cpp
```diff
@@ -52,8 +55,51 @@ namespace rcon
 			network::send(target, "print", console_buffer);
 		}
 
+		bool rate_limit_check(const game::netadr_t& address, const int time)
+		{
+			const auto hash = std::hash<game::netadr_t>()(address);
```
momo5502, line 60: honestly no need to hash, just use the netadr as key of the map. that is supported

momo5502 (changes requested):
looks good, only one minor thing

src/client/component/rcon.cpp
diamante0018: Done

src/client/component/rcon.cpp
```diff
@@ -57,20 +57,19 @@ namespace rcon
 
 		bool rate_limit_check(const game::netadr_t& address, const int time)
 		{
-			const auto hash = std::hash<game::netadr_t>()(address);
-			if (!rate_limit_map.contains(hash))
+			if (!rate_limit_map.contains(address))
 			{
-				rate_limit_map[hash] = 0;
+				rate_limit_map[address] = 0;
 			}
```
momo5502: don't think that's necessary, accessing the map in line 65 will default construct the entry anyways if it's not contained. the default value for int should be zero, so i think you can yeet ll 60-63

momo5502 (changes requested):

src/client/component/rcon.cpp
diamante0018: I had to test it to believe it but it's true

diamante0018: Done

momo5502: at() throws an exception, operator[] default constructs :D

momo5502 (approved):

PR #596, opened 2023-04-25
Fix incorrect mod data in getInfo

Brentdevent:
Forgot to check if `fs_game` is an empty string. 😅 
This fixes non modded private matches again.


diamante0018 (changes requested):
```
if (result.size() > 31)
{
	result.resize(31);
}
```
Should be changed to `result.substr(0, 31);` in this file.

diamante0018 (approved):

momo5502 (approved):

PR #638, opened 2023-05-02
fix: add error message when no BO3 executable is found

JariKCoding:
Gives it a proper error message instead of "Failed to map BlackOps3.exe"

image


momo5502:
very gud

PR #647, opened 2023-05-04
fix: only load the gsc files with the right magic

JariKCoding:


src/client/component/script.cpp
```diff
@@ -69,7 +69,7 @@ namespace script
 			{
 				std::string data;
 				auto script_file = script.generic_string();
-				if (!std::filesystem::is_directory(script) && utils::io::read_file(script_file, &data))
+				if (!std::filesystem::is_directory(script) && utils::io::read_file(script_file, &data) && *(uint64_t*)data.c_str() == 0x1C000A0D43534780)
```
diamante0018, line 72: Feels kinda hacky. We do not use c_str but rather we use data. Then this is a C cast rather than a C++ style cast.
Would suggest breaking this out in a new if statement.
Also you should probably use https://en.cppreference.com/w/cpp/string/byte/memcmp

diamante0018 (changes requested):

src/client/component/script.cpp
diamante0018: also define the magic as a constexpr value somewhere so we can infer by the name that this random number is supposed to be "GSC_MAGIC" or something like that

momo5502: i aggree with the GSC magic. also make sure the endianness is correct. and most importantly data.size() >= sizeof(GSC_MAGIC).

diamante0018 (approved):
Great job

momo5502:
will merge tomorrow

momo5502 (approved):

PR #649, opened 2023-05-05
maint(cryptography): reduce code generation

diamante0018:
by at least 3 asm instructions

😱

### DIFF
--- a/src/common/utils/cryptography.cpp
+++ b/src/common/utils/cryptography.cpp
@@ -240,7 +240,7 @@ namespace utils::cryptography
 			return {cs(buffer), length};
 		}
 
-		return "";
+		return {};
 	}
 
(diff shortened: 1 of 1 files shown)


momo5502:
i like {}. what's the advantage of using "" over {}?

diamante0018:
> i like {}. what's the advantage of using "" over {}?

I am using {} over "" in this PR
The *disadvantage* of using "" is that the compiler optimization may be unable to realize at compile time that strlen("") is 0 when initializing the string object and extra code is generated.
In this PR I am using {} to make sure that does not happen and the compiler calls default string ctor.

diamante0018:
I believe there might have been a misunderstanding where you this I am using "" instead of {} but it is the other way around.

PR #667, opened 2023-05-09
fix: hot join issues for zombies & campaign

diamante0018:


diamante0018:
@JezuzLizard

pistakilla:
👀

src/client/game/symbols.cpp
```diff
@@ -4,9 +4,14 @@
 
 namespace game
 {
-	eModes Com_SessionMode_GetMode()
+	int Com_SessionMode_GetMode()
```
Brentdevent, line 7: Just a question not sure if you know, since you only refactored this code. Is there actually a reason for the reimplementation? Why isn't just the symbol defined in symbols.hpp :p

src/client/component/patches.cpp
```diff
@@ -38,6 +39,17 @@ namespace patches
 
 			game::Com_Error(game::ERROR_SCRIPT_DROP, "%s", buffer);
 		}
+
+		void scr_get_num_expected_players()
+		{
+			if (game::Com_SessionMode_GetMode() == game::MODE_ZOMBIES)
```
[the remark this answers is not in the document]
diamante0018: Thanks for pointing this out. I will add it.

src/client/game/symbols.cpp
diamante0018: My guess is that the function is inlined and it was put in boiii source code. If not I would not know why it was put in the code like this and not in the symbols.

diamante0018: yes please.

Brentdevent: Com_SessionMode_GetMode -> Server: 0x1405002D0 ; Client: 0x1420F6D30
Com_SessionMode_GetGameMode -> Server: 0x1404FFE50 ; Client: 0x1420F68B0

diamante0018: Done. Should be all.
I also realized some patched needed a refactor because the wrong structs were being used.

PR #675, opened 2023-05-09
ZM: Make Super EE reward compatibility opt-in in mods.

### LINKED ISSUE #469: ZM: "Super Easter Egg"/RK5 Start causes issues with `usermaps` and other mods
### Bug Type

Client, Dedicated Server

### BOIII Version

v0.0.1 (Latest)

### Environment (Operating System)

SteamOS

### Reproducibility Rate

High: The bug can (almost) always be reproduced

### Steps to Reproduce

1. Load a `usermap`, or any mod which may modify the client's starting loadout to have more than one weapon (for example, <https://steamcommunity.com/sharedfiles/filedetails/?id=1128166280>).
2. Start the match, notice that the RK5 is still appended, so the player is holding 3 weapons.
3. Acquire a weapon through normal means (off the wall, Mystery Box, etc.)
4. Server identifies that you are holding too many weapons, client gets "Samantha'd".

### Expected Result

BOIII either does not apply the RK5 reward with a mod or `usermap` loaded, or has a guard against adding it to a Player's loadout if they are already holding 2 weapons.

### Actual Result

See step 4 of reproduction, Player has the "Samantha" effect applied to them.

### (Optional) Proof and Additional Information

2mB too large to embed, unfortunately..
https://youtu.be/1Q-dlkEKjTM

NOTE: I marked this as both client and dedicated server misbehavior.. since this seems like a GSC oversight that and is present in both ded. servers hosting a map/mod, and a private game.

### Severity/Priority

Medium

MotoLegacy:
(Fixes #469)
A consequence of boiii's Online Service emulation carrying over to loaded mods was the RK5 starting weapon causing issues with various ZM maps. If a map/mod already had a second starting weapon defined and applied, the `level.super_ee_weapon` would still get appended, and would cause the "Samantha" effect to trigger. Additionally, levels with custom HUDs via LUI or similar may have certain visualizations of their weapon set (e.g. Daybreak's weapon portraits taken from Black Ops 4) would just display something broken in its place.

This change alleviates this by not removing the feature entirely in loaded mods, but instead making it opt-in, via a new level flag, `level.modSupportsSuperEERewards`. This way a mod creator can evaluate themselves if they want their content to support the weapon for reward holders, and may also encourage them to define their own Super EE weapon themselves by changing the value of `level.super_ee_weapon`.

The change modifies `Callback_PlayerConnect()` in `scripts/zm/gametypes/_globallogic_player`. It was evaluated to be the best location for the change, as it is unlikely to be overridden by a mod to unintentionally break the change, as well as being the place that triggers the stat updating that would define whether or not the player has completed all main quests.

The actual change to the script follows:

```c
// MotoLegacy (09 MAY 2023) -- Super Easter Egg Override START
// Mod creators can use this field to opt-into the Super EE start weapon
if (!IsDefined(level.modSupportsSuperEERewards))
    level.modSupportsSuperEERewards = false;

// No rewards if a mod is loaded and mod creator has not permitted RK5 start.
if (GetDvarString("fs_game") != "" && level.modSupportsSuperEERewards == false)
    self.hasCompletedSuperEE = false;
// MotoLegacy (09 May 2023) -- super Easter Egg Override END
```


data/scripts/zm/gametypes/_globallogic_player.gsc_raw
```diff
@@ -0,0 +1,2618 @@
...
+}
+
+function Callback_PlayerConnect()
+{
+	thread notifyConnecting();
+
+	// MotoLegacy (09 MAY 2023) -- Super Easter Egg Override START
+	// Mod creators can use this field to opt-into the Super EE start weapon
+	if (!IsDefined(level.modSupportsSuperEERewards))
+		level.modSupportsSuperEERewards = false;
```
diamante0018, line 53: Please follow 3arc meme GSC convention:
For example this block should be formatted `if ( getDvarString( "...") )` with the spaces .
I suppose the rest of the files already follows this convention.
You only need to edit the raw file.

diamante0018 (changes requested):

MotoLegacy:
@diamante0018 I have resolved (at least, I think... that convention is whack) the Treyarch GSC convention discrepancy.

diamante0018 (approved):

mjkzy (approved):
nice PR

efinst0rm:
Where is this gsc from? If its from modtools it might be out of date compared to what's used in production.

MotoLegacy:
> Where is this gsc from? If its from modtools it might be out of date compared to what's used in production.

We've talked about this in the Discord — waiting on Serious to confirm it's good but from a skim it seems to match production (yes, it is sourced from the Mod Tools release).

MotoLegacy:
Okay, sure thing. For now I will convert this to a draft PR so people know there is a fix ready, and when the GSC PR is merged I will go ahead and clean up these commits and re-implement it using a custom script.

PR #685, opened 2023-05-12
Stats menu: unlock Easter eggs + set master rank

WantedDV:
- added a toggle to unlock all easter eggs
- players can adjust rank within the Master Prestige up to level 1000.
- setting rank now sets you directly on the rank you choose. No more 1xp to the next level.
- setting level to max now lets you prestige right away.

menu


diamante0018 (approved):
I suppose this is a nice fix.
lua looks good

PR #689, opened 2023-05-13
custom GSC loading

mjkzy:


src/client/component/script.cpp
```diff
@@ -51,11 +52,13 @@ namespace script
...
 
 			auto* rawfile = allocator.allocate<game::RawFile>();
-			rawfile->name = name.c_str();
+			rawfile->name = name.data();
 			rawfile->buffer = file_string;
 			rawfile->len = static_cast<int>(data.length());
 
 			loaded_scripts[name] = rawfile;
+
+			game::Scr_LoadScript(game::SCRIPTINSTANCE_SERVER, name.substr(0, name.size() - 4).data()); // .gsc suffix will be readded by Scr_LoadScript
```
momo5502, line 61: what if name.size() < 4?

src/client/component/script.cpp
```diff
@@ -15,6 +15,7 @@ namespace script
 		constexpr size_t GSC_MAGIC = 0x1C000A0D43534780;
 
 		utils::hook::detour db_findxassetheader_hook;
+		utils::hook::detour load_gametype_script_hook;
```
momo5502, line 18: imo we should try to avoid detours whenever possible and use call hooks instead. detours can break if there are rip-relative instructions within the range of the trampoline

mjkzy: yeah, the hook is not really needed since this is only called once. i'll clean it up later - this had just been a rushed thing :D

momo5502:
the pr looks quite good. why is it still draft?

src/client/component/script.cpp
```diff
@@ -50,12 +50,26 @@ namespace script
 				name.erase(i, host_path.length());
 			}
 
+			auto base_name = name;
+			if (utils::string::ends_with(name, ".gsc"))
+			{
+				base_name = name.substr(0, name.size() - 4); // .gsc suffix will be readded by Scr_LoadScript
+			}
+
+			if (base_name.size() < 1)
```
momo5502, line 59:
```suggestion
			if (base_name.empty())

```

mjkzy:
forgot to change it from a draft

src/client/component/script.cpp
```diff
@@ -49,12 +49,26 @@ namespace script
 				name.erase(i, host_path.length());
 			}
 
+			auto base_name = name;
+			if (utils::string::ends_with(name, ".gsc"))
```
diamante0018, line 53: If the extension is not .GSC you should probably return I think

diamante0018 (commented):
I have a feeling this pr will introduce a massive memory leak

The allocator is never freed

diamante0018 (changes requested):
I have a feeling this pr will introduce a massive memory leak

The allocator is never freed

src/client/component/script.cpp
mjkzy: good call idk why i dont just reutrn lmfao

diamante0018:
It's possible the memory leak already exists on prod boiii right now.
I think it's missing a hook in g_shutdown game to free the scripts

diamante0018 (approved):
lgtm

Aziks0:
The file updater nukes everything in the `data` folder, except the ones listed in `https://updater.xlabs.dev/boiii.json`, so you should whitelist the `data/custom-scripts` folder.

And btw, you used snake cases for `lookup_tables` and `ui_scripts`, so I guess `custom_scripts` might be a better name.

diamante0018:
> The file updater nukes everything in the `data` folder, except the ones listed in `https://updater.xlabs.dev/boiii.json`, so you should whitelist the `data/custom-scripts` folder.
> 
> And btw, you used snake cases for `lookup_tables` and `ui_scripts`, so I guess `custom_scripts` might be a better name.

No. The folder is reversed for your use.
You should use the boiii folder

diamante0018 (approved):

momo5502 (changes requested):

src/client/component/game_event.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include "game_event.hpp"
+
+#include <utils/hook.hpp>
+
+namespace game_event
+{
+	namespace
+	{
+		std::vector<std::function<void()>> g_init_game_tasks_;
```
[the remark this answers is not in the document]
diamante0018: guaranteed yes

src/client/game/structs.hpp
```diff
@@ -1636,6 +1648,44 @@ namespace game
...
+		__int16 vehicleNodeIndex;
+		unsigned __int16 absDynEntIndex;
+		Weapon weapon;
+		uint64_t val;
+	};
+
+	struct scr_entref_t
+	{
+		EntRefUnion u;
+		unsigned __int16 classnum;
```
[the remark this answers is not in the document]
diamante0018: done

src/client/component/game_event.cpp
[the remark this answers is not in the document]
momo5502: as you are providing it as a generic api through a separate component, you can not guarantee it. please use a concurrency container

momo5502 (changes requested):
concurrency container and it seems fine to me

momo5502:
> > concurrency container and it seems fine to me
> 
> why? there should not be any need. functions are always called from game thread.

as i already said, you are exposing this as an api. anyone can include game_event.hpp and access that api. maybe right now you can guarantee thread safety. but anyone using that in the future can suffer from issues. and we will have a hard time debugging. as you decided to model the game events as a general api, we should guarantee thread safety, in my opinion.

momo5502 (approved):

momo5502:
skipping the ci for the last commit was not the best idea, but i'll just force merge and we pray it compiles :P

src/client/component/script.cpp
```diff
@@ -66,31 +89,45 @@ namespace script
...
-			loaded_scripts = {};
 			const utils::nt::library host{};
-			load_scripts_folder((game::get_appdata_path() / "data/scripts").string());
-			load_scripts_folder((host.get_folder() / "boiii/scripts").string());
+
+			const auto data_folder = game::get_appdata_path() / "data";
+			const auto boiii_folder = host.get_folder() / "boiii";
+
+			// scripts folder is for overriding stock scripts the game uses
+			load_scripts_folder((data_folder / "scripts").string(), false);
```
ineedbots, line 125: you forgot gamemode and map specific folders 😭

diamante0018: Noob add them

mjkzy: we just didn't care to add them for the time being but we'll get around to it eventually

diamante0018: You actually left a review. SMH

PR #690, opened 2023-05-13
fix(server_status): format it properly

diamante0018:
bot0 required for consistency between our clients and plutonium as the bot's guid is usually bot%d (%d replaced by qport)
This is not possible currently so just like I've done on plutonium t5 it's just bot0
Then i fixed the status header

### DIFF
--- a/src/client/component/status.cpp
+++ b/src/client/component/status.cpp
@@ -37,7 +37,7 @@ namespace status
 			utils::hook::call(game::select(0x142246EDE, 0x14052C5CE), print_client_xuid);
 
 			utils::hook::copy_string(game::select(0x143050480, 0x140E85A20),
-			                         "num score ping guid             name             address                  qport  \n");
+			                         "num score ping xuid             name             address                  qport  \n");
 			utils::hook::copy_string(game::select(0x1430504E0, 0x140E85A80),
 			                         "--- ----- ---- ---------------- ---------------- ------------------------ ------ \n");
 		}
--- a/src/client/component/status.cpp
+++ b/src/client/component/status.cpp
@@ -1,0 +1,47 @@
+#include <std_include.hpp>
+#include "loader/component_loader.hpp"
+#include <game/game.hpp>
+
+#include <utils/hook.hpp>
+
+namespace status
+{
+	namespace
+	{
+		int client_num_;
+
+		void print_client_num(int channel, int label, const char* fmt, const int client_num)
+		{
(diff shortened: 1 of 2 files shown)


src/client/component/status.cpp
```diff
@@ -0,0 +1,47 @@
...
+	struct component final : generic_component
+	{
+		void post_unpack() override
+		{
+			// Patch the status command for test clients
+			utils::hook::call(game::select(0x142246E37, 0x14052C527), print_client_num);
+			utils::hook::call(game::select(0x142246EDE, 0x14052C5CE), print_client_xuid);
+
+			utils::hook::copy_string(game::select(0x143050480, 0x140E85A20),
+			                         "num score ping guid             name             address                  qport  \n");
```
momo5502, line 40: original status says xuid, doesn't it?

diamante0018: our clients have guid. storm asked me to rename but you would be correct. the original status says xuid

momo5502: note there is two status handlers in the client, one that says guid and the one that says xuid. i think the one with xuid is the one that's actively used. so make sure you base this on the proper on

diamante0018: I made sure this is the one that prints into the console when i type status command in my in-game console

diamante0018: i renamed it back to xuid so it stays consistent. but this should be the correct status handler between the two

momo5502: perfect, just wanted to make sure.

momo5502 (approved):

src/client/component/status.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "loader/component_loader.hpp"
+#include <game/game.hpp>
+
+#include <utils/hook.hpp>
+
+namespace status
+{
+	namespace
+	{
+		int client_num_;
```
momo5502, line 11: i recommend this to be thread_local

momo5502 (changes requested):

src/client/component/status.cpp
momo5502: even though technically it's not really necessary,calling the status command from different threads concurrently could yield errors. we know this is not gonna happen, but using thread_local still seems to be a good practice in that case

diamante0018: done

momo5502 (approved):

PR #698, opened 2023-05-13
fix: better scr_get_num_expected_players patch

diamante0018:
This fixes the patch and removes the very hot fix I had to push to avoid game's breaking on private match.


diamante0018:
This should fix the patch for good. My bad.

momo5502 (approved):

PR #703, opened 2023-05-14
Restores game events to games_mp log

efinst0rm:
Remove dev blocks from logprint function.
Sets bots xuid's as "bot0"


diamante0018 (approved):

metalglove:
When logging a chat event like `Heey! ;)`, wouldn't that invalidate the chat log line or player name? Or at least cut it off at the end. Shouldn't player names and chat messages be put into quotes or something? i.e. ("name "SuperSmileyMan ;-)" and chat message "Heey! ;))

I see that IW4Admin already does this wrong.
image
Only prints the `)` character.
image

diamante0018:
The call to print chat messages is done from the engine. It is unrelated to this GSC file

momo5502:
can/should thiss be merged? st0rm asked me to close this, so idk what's the state?

efinst0rm:
> can/should thiss be merged? st0rm asked me to close this, so idk what's the state?

Good to go!

JezuzLizard:
IMO there is a better way to go about this now that we have custom script loading. Replacing the entire file makes sense on IW6x and S1x because there was(still is) no modding scene for conflicts to be a concern. However T7 has had mod tools for almost 7 years, and as a result 1000s of mods have been created, yes mostly for ZM but MP has had mods made for it too.
Due to the lack of replaceFunc() modders have to replace entire files if they need to modify even a single function in that file.  Because of this, overriding GSC files has to be done carefully to minimize the likelihood of breaking mods because boiii's GSC's have the highest priority they will overwrite the mod's GSCs.

Anyway from what I see with the changes made to the file to restore the logprints you can make a custom script that adds a connect callback(to restore the connect logprint), and disconnect callback(to restore the disconnect logprint). Restoring the damage logprint is kind of difficult right now until replaceFunc() is added because while you could set the callback_onplayerdamage function pointer to your own it would break mods that modify that function, and that function is fairly common to edit for various reasons. Lastly to do the logprint for killed players I would recommend storing the original callback_playerkilled function in a level variable, set the level.callbackplayerkilled to your own function, do the logprint and then call the original at the end.

All in all there is no reason to replace the entire file in this case. i think @mjkzy could add replaceFunc() if requested and that would allow you to do the damage logprint.

I think we can all agree that we should avoid emulating IW4x when it comes to mod compatibility. It seems that every week some change is made to IW4x like removing a builtin or adding a builtin that conflicts with mod code or some other change that breaks mods. The modding scene in T7 is much bigger than IW4x's so it would be even more detrimental to break mods unnecessarily.

momo5502:
> 

good point. i'm still going to merge this, becasue if a mod overrides this gsc file, it breaks the game_log functionality added by this script, and nothing more (to my understanding). but in general, we should really follow your philosophy. i totally agree

PR #714, opened 2023-05-16
load our scripts sooner

mjkzy:
fixes a problem where bots weren't spawning in combat training (due to original _bot.gsc loading before our override?)
image

using this patch, bots do indeed spawn again
image

### DIFF
--- a/src/client/component/script.cpp
+++ b/src/client/component/script.cpp
@@ -159,14 +159,12 @@ namespace script
 
 		void begin_load_scripts_stub(game::scriptInstance_t inst, int user)
 		{
-			if (!game::Com_IsInGame() || game::Com_IsRunningUILevel())
-			{
-				game::Scr_BeginLoadScripts(inst, user);
-				return;
-			}
-
 			game::Scr_BeginLoadScripts(inst, user);
-			load_scripts();
+
+			if (game::Com_IsInGame() && !game::Com_IsRunningUILevel())
+			{
+				load_scripts();
+			}
 		}
 
 		int server_script_checksum_stub()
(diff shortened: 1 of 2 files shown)


src/client/component/script.cpp
```diff
@@ -157,15 +157,15 @@ namespace script
...
 			{
-				game::GScr_LoadGametypeScript();
+				game::Scr_BeginLoadScripts(inst, user);
 				return;
 			}
 
-			game::GScr_LoadGametypeScript();
+			game::Scr_BeginLoadScripts(inst, user);
 			load_scripts();
 		}
```
momo5502, line 170:
```C++
		void begin_load_scripts_stub(game::scriptInstance_t inst, int user)
		{
			game::Scr_BeginLoadScripts(inst, user);

			if (game::Com_IsInGame() && !game::Com_IsRunningUILevel())
			{
				load_scripts();
			}
		}
```

momo5502 (changes requested):

src/client/component/script.cpp
mjkzy: ah yeah, i could just do this instead

src/client/game/symbols.hpp
```diff
@@ -243,8 +244,6 @@ namespace game
 		0x1422E9410, 0x1405811E0
 	};
 
-	WEAK symbol<void()> GScr_LoadGametypeScript{0x141AAD850, 0x1402D7140};
-
```
diamante0018, line 247: I don't think you need to delete the old symbol even if unused. Still handy to have for people starting to reverse the game I guess.
You don't need to add it back in my opinion. Just keep in mind for the next time

diamante0018 (approved):

momo5502 (approved):

PR #717, opened 2023-05-16
feat: custom players folder

### LINKED ISSUE #394: having it's own "players" folder.
### Duplicate Issue Check

- [X] I have searched the list of issues on this repository. I confirm that this is not a duplicate of any open or closed issue(s).

### Your feature request or suggestion:

Seems a conflict going on between steam and boiii if you using the same game files folder. If you change your settings too much. You would not be able to play normal T7 matchmaking. you would get kick saying you been kicked from Call of Duty Servers. 

Sadly this will confused everyone right now with loss of stats. so best to wait for official release?

diamante0018:
Closes #394

### DIFF
--- a/src/client/component/client_patches.cpp
+++ b/src/client/component/client_patches.cpp
@@ -4,7 +4,9 @@
 #include "scheduler.hpp"
 
 #include <game/game.hpp>
+
 #include <utils/hook.hpp>
+#include <utils/io.hpp>
 
 #include <mmeapi.h>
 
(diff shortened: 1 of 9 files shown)


diamante0018:
I might have missed one.

momo5502 (changes requested):
code looks good, but two things:

we use players internally as well, for example for demonware user files, or for the dvar config and a few other things i think. replacing those occurances should be done as well imo.

also, there should be a small migration process that for example checks if `players` exists, but not `boiii_players` and then copies over the folder. that way people who are playing right now don't lose their stats. we can remove that code in a week or two then.

diamante0018:
I tested the migration code and it works. Decided to execute it before the game runs for good effect.

momo5502 (approved):

PR #748, opened 2023-05-20
fix(scripts): fix for me being a bit silly

diamante0018:
Oopsie


diamante0018:
Big fix, forgive me momo

PR #754, opened 2023-05-21
ZM: Make Super EE reward compatibility opt-in in mods.

### LINKED ISSUE #469: ZM: "Super Easter Egg"/RK5 Start causes issues with `usermaps` and other mods

MotoLegacy:
This is a re-draft of #675 as reverting the previous commits to clean the Git history had the consequence of the PR being auto-closed. This PR fixes #469.

A consequence of boiii's Online Service emulation carrying over to loaded mods was the RK5 starting weapon causing issues with various ZM maps. If a map/mod already had a second starting weapon defined and applied, the `level.super_ee_weapon` would still get appended, and would cause the "Samantha" effect to trigger. Additionally, levels with custom HUDs via LUI or similar may have certain visualizations of their weapon set (e.g. Daybreak's weapon portraits taken from Black Ops 4) would just display something broken in its place.

This change alleviates this by not removing the feature entirely in loaded mods, but instead making it opt-in, via a new level flag, `level.modSupportsSuperEERewards`. This way a mod creator can evaluate themselves if they want their content to support the weapon for reward holders, and may also encourage them to define their own Super EE weapon themselves by changing the value of `level.super_ee_weapon`.

The previous Pull Request modified and overrode `scripts/zm/gametypes/_globallogic_player.gsc`. This new one uses the custom GSC feature to create a new script, `super_ee_starting_weapon.gsc`.


data/custom_scripts/zm/super_ee_starting_weapon.gsc_raw
```diff
@@ -0,0 +1,38 @@
...
+
+    for ( ;; )
+    {
+        level waittill( "connecting", player );
+
+        // Force-set the completion stat to false if prohibited.
+        if ( isDefined( disallow_super_ee_rewards ) )
+            player.hasCompletedSuperEE = false;
+    }
+}
```
diamante0018, line 38: Bo3 GSC scripts have a built-in callback for connecting players
You should use that.

diamante0018 (changes requested):

data/custom_scripts/zm/super_ee_starting_weapon.gsc_raw
MotoLegacy: Does that have the potential to conflict with mods though? I had looked at alternatives and everything regarding callbacks I had found seemed to be that case -- the functionality looked like it would break a mod. Please direct me to what you're referring to otherwise.

diamante0018:
```
/@
"Name: on_connect(<func>)"
"Module: Callbacks"
"Summary: Set a callback for when a player connects"
"MandatoryArg: <func> the function you want to call on the new player."
"Example: callback::on_connect(&on_player_connect);"
"SPMP: singleplayer"
@/
function on_connect( func, obj )
{
	add_callback( #"on_player_connect", func, obj );
}
```
You will probably use this

